PR #1, opened 2019-07-31
Update README to consistently use yarn

asheemmamoowala:


mourner:
I'm not sure if the docs team wants to continue using Yarn — now that the project is independent, we could switch to NPM (and package lock) if it's more convenient for the team.

colleenmcginnis:
>I'm not sure if the docs team wants to continue using Yarn — now that the project is independent, we could switch to NPM (and package lock) if it's more convenient for the team.

Yes. I think we will want to switch to npm. We also may want to use `npm start` instead of `npm run start-docs` now that this is a separate project. That matches patterns in other docs sites -- we  want to make this repo consistent with other repos wherever we can (for easier maintenance on the docs side) without disrupting your existing workflows too much. 🙂

mourner:
Closing in favor of https://github.com/mapbox/mapbox-gl-js-docs/commit/09b2ac3045c04952e822822efac05d5e70465807

PR #5, opened 2019-08-01
[staging] Set up for publisher

colleenmcginnis:
Same as #4 but for the staging stack.


mourner:
I just rebranched `publisher-staging` after merging the previous PR so no need to merge separately.

PR #22, opened 2019-08-13
add metadata for using text-color within formatted expressions

ansis:
The docs-side change of https://github.com/mapbox/mapbox-gl-js/pull/8068


chloekraw:
hmm, how can we clearly communicate in our docs that using `text-color` in a `formatted` expression is only supported starting GL-JS SDK version 1.3?

PR #23, opened 2019-08-14
update example to use MercatorCoordinate meterInMercatorCoordinateUnits method

andrewharvey:
corresponding docs update from https://github.com/mapbox/mapbox-gl-js/pull/8524

Need to wait for https://github.com/mapbox/mapbox-gl-js/pull/8524 to land in the next release before we can merge this.


andrewharvey:
I've created a label to help organise PRs which can be merged when ready and PRs which can't be merged as they depend on a feature which will become available in a future GL JS release. Feel free to rename the label, but I do think we should have at least some kind of label.

We could build it into the GL JS release instructions to go through here and merge any approve and "next release" PRs.

PR #57, opened 2019-08-26
remove eslint comments from examples

### LINKED ISSUE #56: Eslint comments in example code
https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-draw/
https://github.com/mapbox/mapbox-gl-js-docs/blob/publisher-production/docs/pages/example/mapbox-gl-draw.html#L30

https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder-accept-coordinates/
https://github.com/mapbox/mapbox-gl-js-docs/blob/publisher-production/docs/pages/example/mapbox-gl-geocoder-accept-coordinates.html#L31-L32

There may be more...

katydecorah:
fixes #56


docs/pages/example/mapbox-gl-geocoder-accept-coordinates.html
```diff
@@ -28,8 +28,8 @@
                 type: "Point",
                 coordinates: [lng, lat]
             },
-            place_name: 'Lat: ' + lat + ' Lng: ' + lng, // eslint-disable-line camelcase
-            place_type: ['coordinate'], // eslint-disable-line camelcase
+            place_name: 'Lat: ' + lat + ' Lng: ' + lng,
+            place_type: ['coordinate'],
```
andrewharvey: I agree these shouldn't be in the published example, but I am curious why eslint would no longer complain about these if we remove the eslint-disable.

PR #88, opened 2019-09-19
Upgrade Sentry

katydecorah:
* [x] install @sentry/browser: `npm install @sentry/browser`
* [x] in page-shell.js:
    * [x ] add `import * as Sentry from '@sentry/browser';` to top of page
    * [x] remove the `sentry` object from `window.initializeMapboxAnalytics()` in page-shell.js
    * [x] add `Sentry.init({ dsn: 'https://6ba8cfeeedad4fb7acb8576f0fd6e266@sentry.io/1384508' });` to `componentDidMount()` in page-shell.js
* [x] run `npm start`, check console for unusual errors/warnings
* [x] test Sentry, reviewers can perform this check as well:
    * [x] temporarily add `Sentry.captureException(new Error("Upgrading Sentry"));` to the page:
    * [x] run `npm start` and visit page, check Sentry for the event
    * [x] open site in IE11, check Sentry for the event
    * [x] remove test code

Ref https://github.com/mapbox/documentation/issues/267


docs/components/page_shell.js
```diff
@@ -28,23 +29,11 @@ class PageShell extends React.Component {
     componentDidMount() {
         // initialize analytics
         if (typeof window !== 'undefined' && window.initializeMapboxAnalytics) {
-            const isProduction = /\.?mapbox\.com/.test(
```
danswick: Do you think it's worth tagging with the environment still? https://docs.sentry.io/enriching-error-data/environments/?platform=browser#how-to-send-environment-data

Or is that something we should considering going back to add to all the docs sites later?

katydecorah: I was considering adding it, but I think it makes sense to filter out all localhost/staging events since we don't do anything with that data right now. Definitely something we can revisit and add later!

PR #105, opened 2019-10-10
Update image expression docs

ryanhamley:
Adds an `image` expression operator example and ensures the operator shows up in the docs


docs/pages/style-spec.js
```diff
@@ -1348,14 +1348,31 @@ export default class extends React.Component {
...
+                        <SectionH3
+                            id="types-resolved-image"
+                            title="ResolvedImage"
+                        >
+                            <p>
+                                The <code>ResolvedImage</code> type represents
+                                an image which will be checked against the
+                                current style to see if it is available or not.
+                                This allows for complex image logic such as
+                                fallback and default images.
```
kkaefer: Could you please explain the difference between a plain string, and a resolved image? The text is still a bit unclear about why you'd want to use an `image` type instead of a plain string.

docs/data/style-spec-navigation.js
```diff
@@ -144,6 +144,9 @@ export const styleSpecNavigation = [
             {
                 title: 'Formatted'
             },
+            {
+                title: 'Resolved Image'
```
kkaefer: The heading is titled `ResolvedImage`, while the sidebar uses `Resolved Image`. Let's standardize on one spelling.

docs/pages/style-spec.js
asheemmamoowala: In this context, there is no difference between `fallback` and `default` images.

ryanhamley:
I think this will have to wait for https://github.com/mapbox/mapbox-gl-js/pull/8841 to land first

ryanhamley:
Tagging @chloekraw for a review on language. Also, @colleenmcginnis because I had to monkey patch the code that creates the sidebar URLs to work correctly with a two word type (see the changes in `page_shell.js`) and wanted to make sure that was ok or I wasn't missing an easier way to do it. Basically, I wanted the type to be `resolvedImage` which correctly produces the slug `#type-resolvedImage` from all the links except the sidebar which gets lowercased to `#type-resolvedimage`.

docs/pages/style-spec.js
```diff
@@ -1348,14 +1348,35 @@ export default class extends React.Component {
                                     }`)}
                         </SectionH3>
 
+                        <SectionH3
+                            id="types-resolvedImage"
+                            title="ResolvedImage"
+                        >
+                            <p>
+                                The <code>resolvedImage</code> type represents
```
chloekraw: This sounds good. I might recommend explicitly connecting "image" to "icon or pattern" by:

> The <code>resolvedImage</code> type represents an image that exists because it is used on a layer, e.g., an icon or pattern.

(adapted from @ansis' explanation in slack about the difference between an icon and an image. we should make sure this is accurate before committing)

docs/pages/style-spec.js
```diff
@@ -1348,14 +1348,35 @@ export default class extends React.Component {
...
+                                an icon or pattern. The resolvedImage is checked
```
chloekraw: This sentence a bit ambiguous, and I'm left wondering what/when/how the checking described is actually happening. Here's a stab at making this clearer if my understanding is correct:

> An input to the `image` expression operator is checked against the current map style to see if it is available to be rendered or not, and the result is returned in the `resolvedImage` type.

docs/pages/style-spec.js
```diff
@@ -1348,14 +1348,35 @@ export default class extends React.Component {
...
+                                against the current map style to see if it is
+                                available to be rendered or not. This approach
+                                allows developers to define a series of images
+                                which the map can fall back to if previous
+                                images are not found, which cannot be achieved
+                                by providing, for example, `icon-image` with a
+                                plain string.
```
chloekraw: nit: explain why it's not possible to do this with `icon-image` to make this point complete

docs/pages/style-spec.js
```diff
@@ -1354,14 +1354,22 @@ export default class extends React.Component {
...
+                                in the
+                                <code>resolvedImage</code> type. This approach
                                 allows developers to define a series of images
                                 which the map can fall back to if previous
                                 images are not found, which cannot be achieved
-                                by providing, for example, `icon-image` with a
-                                plain string.
+                                by providing, for example,
+                                <code>icon-image</code> with a plain string
+                                (because you cannot define multiple images in a
```
chloekraw: nit: rewrite to avoid second person (not sure if it's used elsewhere in our docs, but it currently isn't in this entry)

> (because multiple strings cannot be supplied to `icon-image` and multiple images cannot be defined in a string)

docs/pages/style-spec.js
```diff
@@ -1354,14 +1354,22 @@ export default class extends React.Component {
                         >
                             <p>
                                 The <code>resolvedImage</code> type represents
-                                an icon or pattern. The resolvedImage is checked
-                                against the current map style to see if it is
-                                available to be rendered or not. This approach
+                                an image (e.g., an icon or pattern) which is
+                                used in a layer. An input to the{' '}
```
chloekraw: what is the `{' '}`?

ryanhamley: Looks like it's added automatically during a commit check. I presume it tells the HTML builder to add a space or something. This is used in a bunch of other places in this file and the docs render properly

colleenmcginnis: ☝️ that's right @ryanhamley.

PR #109, opened 2019-10-10
Update GL JS to version v1.4.1

ryanhamley:


HeyStenson:
@ryanhamley I've never approved a PR that's only submodule changes before! 😅 Anything in particular I should be on the lookout for?

ryanhamley:
Not really. You can check the `package.json` to see that it's the correct version (1.4.1) https://github.com/mapbox/mapbox-gl-js/compare/7aa025af1e71027d3d17f2ca341e835acc3b63e2...81e333e9575f214c19b6253742e94d9f1071f81b#diff-b9cfc7f2cdf78a7f4b91a753d10865a2 but I think that @colleenmcginnis added a check to make sure the submodule can't be moved backwards and since it wouldn't show any changes at all if the submodule were the same, I guess if you see changes it has to be because it was updated to the new version 🤷‍♂

PR #115, opened 2019-10-16
mapbox-gl-geocoder@v4.5.0

scottsfarley93:
Upgrades mapbox-gl-geocoder to the latest version (v4.5.0) (\ref https://github.com/mapbox/mapbox-gl-geocoder/pull/303)

Also uses the new `addTo` method on the geocoder to demonstrate how to add the geocoder to a non-map html element (\ref https://github.com/mapbox/mapbox-gl-geocoder/pull/295). 

\cc @mapbox/docs


katydecorah:
@scottsfarley93 when I tested /mapbox-gl-js/example/mapbox-gl-geocoder-outside-the-map/ by entering a search and then clicking an item from the autocomplete, nothing happens (map does not move to the location). There's no errors logged to the console either 🤔

scottsfarley93:
Interesting @katydecorah -- thanks for catching this. I need to double check to confirm, but I'm pretty sure this is the intended behavior. 

Since this example page is showing how to put an geocoder into an arbitrary non-map html element, I thought it would be good to demonstrate using the new functionality introduced in v4.5. I'm not realizing that this is probably not a good example anymore, because, as you point out, the point of this example is to maintain the connection between map and geocoder -- just without having the geocoder inside of the map's html. 

I can revert this particular example back to the way it was and perhaps add a new example of using the geocder entirely without a map.

katydecorah:
> I can revert this particular example back to the way it was and perhaps add a new example of using the geocder entirely without a map.

@scottsfarley93 I think this is a good plan!

scottsfarley93:
Created the new example. @katydecorah I'm having some trouble generating the image associated with the new (no map) example. Would you mind giving it a look over when you get a chance?

docs/pages/example/mapbox-gl-geocoder-no-map.html
```diff
@@ -0,0 +1,36 @@
...
+    }
+    .mapboxgl-ctrl-geocoder {
+        min-width: 100%;
+    }
+</style>
+<div id="geocoder"></div>
+<script>
+    var geocoder = new MapboxGeocoder({
+        accessToken: mapboxgl.accessToken,
+        types: [
```
katydecorah: @scottsfarley93 The screenshot is rendering blank because this example has a small error here. `types` should be a comma delimitated string.

After you fix that try running `npm run  create-image 3d-buildings` again!

docs/pages/example/point-from-geocoder-result.html
```diff
@@ -1,9 +1,18 @@
+<<<<<<< HEAD
```
katydecorah: Looks like this merge conflict snuck in

docs/pages/example/forward-geocode-custom-data.html
```diff
@@ -1,9 +1,18 @@
+<<<<<<< HEAD
```
katydecorah: Another merge conflict we'll want to resolve!

katydecorah:
@scottsfarley93 left you a few comments to help with that image! You'll also want to update from the latest on publisher-production again to ease the warning about the submodule.

scottsfarley93:
Thanks for having a look @katydecorah! I added the new example, so I think this change is good to go.

andrewharvey:
@scottsfarley93 I think this just needs a rebase against publisher-production and at the same time we can upgrade to `v4.5.1`

PR #122, opened 2019-10-28
add essential: true option for animations considered essential with respect to prefers-reduced-motion

### LINKED ISSUE #117: Feedback for /mapbox-gl-js/example/flyto-options/
*This feedback was submitted by a user through the feedback widget on the bottom of the page.*

* Page: [/mapbox-gl-js/example/flyto-options/](https://docs.mapbox.com/mapbox-gl-js/example/flyto-options/)
* Section: null
* Date received: 2019-10-17T10:09:27.783Z
* Was this page helpful?: 😕

Feedback:
> The example is not working...there's no "slowly fly"

This is the second time this week we've gotten feedback on this example saying that it's not working.  (The other feedback, from 10/12, was simply: `It doesnt work`.) As far as I can tell, the example does exactly what it's supposed to - 

> // These options control the flight curve, making it move
> // slowly and zoom out almost completely before starting
> // to pan.  

However, the fact that we've gotten two separate pieces of feedback about this makes me think that user expectations may be different. What can we clarify in either the comments or the example header?

andrewharvey:
closes #117

not to be merged until when or if https://github.com/mapbox/mapbox-gl-js/pull/8883 is released


andrewharvey:
pending a review this can be merged after #164

andrewharvey:
> Works great!

@katydecorah Thanks, I can't merge here so if you could when you're ready please.

PR #124, opened 2019-10-29
Add tape test to verify pinned version of GL-JS is the latest stable

danswick:
To help ensure contributors don't accidentally change the pinned version of the Mapbox GL JS submodule, this PR adds a new test that compares the currently pinned version to the latest stable tag in the Mapbox GL JS repo. 

A couple of notes: 

- While testing, I noticed that, because there are multiple packages in the Mapbox GL JS repo, multiple tags can sometimes be associated with the same SHA. Because of this, we need a bit of extra logic to find all of the tags associated with a given ref and filter out any tags that are prereleases or related to the Style Spec before comparing to the current tag.
- I'm not sure if there are any scenarios where we'll want the Mapbox GL JS version to be on anything other than the most recent stable tag. If those scenarios exist, we'll need to make some adjustments here.
- This PR doesn't attempt to address the versioning expectations for example generation. We may want to address that as well before merging. 

- Sample failing test: https://travis-ci.com/mapbox/mapbox-gl-js-docs/builds/134053334
- Sample passing test: https://travis-ci.com/mapbox/mapbox-gl-js-docs/builds/134056268


danswick:
Great suggestion, @asheemmamoowala. Made that change and it seems to be working great.

test/latest-version.test.js
```diff
@@ -0,0 +1,32 @@
...
+        tags.push(`Current tag: ${current}`);
+        return false;
+    }
+}
+
+test(`Is latest version`, t => {
+    tagsMatch().then(res => {
+        t.ok(
+            res,
+            `Mapbox GL JS submodule is not pinned to the latest stable release version. ${tags}`
```
katydecorah: What do you think about adding an action to this error? If I received this error, I'm not sure how to fix it.

Issue #100, opened 2019-10-02
Feedback for /mapbox-gl-js/example/add-image-animated/


danswick:
*This feedback was submitted by a user through the feedback widget on the bottom of the page.*

* Page: [/mapbox-gl-js/example/add-image-animated/](https://docs.mapbox.com/mapbox-gl-js/example/add-image-animated/)
* Section: null
* Date received: 2019-10-01T15:54:01.080Z
* Was this page helpful?: 😕

Feedback:
> Some comments to explain what exactly is going on?  When is onAdd() called?  When is render() called?

This example is quite confusing. It's not clear that the image being passed to `addImage` is being drawn. And it's not clear why it's being done that way.

asheemmamoowala:
The goal of this example is to document how to custom draw an icon. The `render` method is called for each frame, and since it calls `map.triggerRepaint`, it is repainting the map continuously, resulting in the smooth animation of the dot.

cc @ansis

PR #133, opened 2019-11-01
Add comments to add-image-animated example

adrianababakanian:
Fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/100


docs/pages/example/add-image-animated.html
```diff
@@ -9,18 +9,21 @@
 
 var size = 200;
 
+// custom drawn pulsing dot icon
```
ryanhamley: i think we should probably call out that this is an implementation of the CustomLayerInterface

maybe something like 

```
// implementation of CustomLayerInterface to draw a pulsing dot icon on the map
// see https://docs.mapbox.com/mapbox-gl-js/api/#customlayerinterface for more info
```
?

docs/pages/example/add-image-animated.html
```diff
@@ -9,18 +9,21 @@
...
 var pulsingDot = {
     width: size,
     height: size,
     data: new Uint8Array(size * size * 4),
 
+    // get drawing context for icon once it is added to the map
```
ryanhamley: `// get rendering context for the map canvas when layer is added to the map`

docs/pages/example/add-image-animated.html
```diff
@@ -48,7 +51,7 @@
         // update this image's data with data from the canvas
         this.data = context.getImageData(0, 0, this.width, this.height).data;
 
-        // keep the map repainting
+        // keep repainting the map continuously, resulting in the smooth animation of the dot
```
ryanhamley: could simplify this a bit to just `continuously repaint the map, resulting in the smooth animation of the dot`

docs/pages/example/add-image-animated.html
ryanhamley: or just call it a custom style layer...either way gives them more context to search the docs for

adrianababakanian: I agree that calling out `CustomLayerInterface` and linking the relevant section in the API reference is best for docs search-ability.

Issue #111, opened 2019-10-11
Feedback for /mapbox-gl-js/example/geojson-markers/


HeyStenson:
*This feedback was submitted by a user through the feedback widget on the bottom of the page.*

* Page: [/mapbox-gl-js/example/geojson-markers/](https://docs.mapbox.com/mapbox-gl-js/example/geojson-markers/)
* Section: null
* Date received: 2019-10-10T17:40:46.602Z
* Was this page helpful?: 😕

Feedback:
> Where in god's name are those icons defined?

I think this demonstrates again that in-line comments could be really useful in these examples. Unless someone is really familiar with GL JS syntax, it's understandable that they aren't clear on how the two icons are being imported.

PR #134, opened 2019-11-01
Add inline comments for geojson-markers example

adrianababakanian:
Fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/111


docs/pages/example/geojson-markers.html
```diff
@@ -40,12 +42,15 @@
             }
         },
         "layout": {
+            // name of image in the style's sprite, where `icon` is pulled from properties.icon
             "icon-image": "{icon}-15",
```
ryanhamley: I would suggest that this is a good occasion to update this example to remove this old token syntax and replace it with a more up to date expression

so instead of 

`"icon-image": "{icon}-15"`

it should be

`"icon-image": ["concat", ["get", "icon"], "-15"]`

ryanhamley: In this case, the comment could be something like

```
// get the icon name from the source's "icon" property
// concatenate the name to get an icon from the style's sprite sheet
```

docs/pages/example/geojson-markers.html
```diff
@@ -40,12 +42,15 @@
...
+            // value to use for text label, where `title` is pulled from properties.title
             "text-field": "{title}",
```
ryanhamley: `"text-field": ["get", "title"]`

adrianababakanian: Thanks for catching this!

docs/pages/example/geojson-markers.html
```diff
@@ -40,12 +42,16 @@
             }
         },
         "layout": {
-            "icon-image": "{icon}-15",
-            "text-field": "{title}",
+            // get the icon name from the source's "icon" property
+            // concatenate the name to get an icon from the style's sprite sheet
+            "icon-image": ["concat", ["get", "icon"], "-15"],
+            // get the title name from the source's "tilte" property
```
ryanhamley: typo: `tilte`

adrianababakanian: Fixed 👍

Issue #89, opened 2019-09-20
'Get features under mouse pointer' performs poorly on Android Chrome


asheemmamoowala:
The example page at https://docs.mapbox.com/mapbox-gl-js/example/queryrenderedfeatures/, takes 7-10s to update on Android Chrome (running on a Oneplus 5t)

Debugging in Chrome shows that much of the time is spent in the `HTML Layout` step when updating the `innerHTML` of the features overlay

https://github.com/mapbox/mapbox-gl-js-docs/blob/97c75229d0d953bcea7f06af7544ccf403e9d4fa/docs/pages/example/queryrenderedfeatures.html#L27

One way to address this might be to display fewer attributes of the features (excluding the geometry, for instance)

PR #142, opened 2019-11-14
Fix example queryrenderedfeatures

ahk:
Fixes two small issues with the `queryrenderedfeatures` example page.
1. Performance on mobile phones (specifically tested Chrome android)
2. Size of output panel was fixed width (also annoying on mobile).

Fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/89


docs/pages/example/queryrenderedfeatures.html
```diff
@@ -24,6 +24,27 @@
 
 map.on('mousemove', function (e) {
     var features = map.queryRenderedFeatures(e.point);
-    document.getElementById('features').innerHTML = JSON.stringify(features, null, 2);
+
+    // Limit the number of properties we're displaying
+    // because the full set was very slow to display and too verbose to read
```
danswick: I might reword this a bit. I don't think folks need to know about the example from the past or that it would otherwise be slow. Maybe something like: 

> Limit the number of properties to make the list easier to read

ahk: I made it simpler and removed reference to the past, but I think it's good for people to realize this could be a performance issue, since unacceptable performance on mobile is why I'm updating the example.

Issue #139, opened 2019-11-12
Embed live code snippets in GL JS examples


adrianababakanian:
## Motivation
Currently, the flow for working off of sample code in the [Mapbox GL JS examples page](https://docs.mapbox.com/mapbox-gl-js/examples/) looks something like the following:

1. Creating an empty HTML file.
2. Copy-pasting the example into the new file. 
3. Making adjustments to customize the example from there.

## Proposal
Creating a simple embedded IDE (similar to [JSFiddle](https://jsfiddle.net/), [JS Bin](https://jsbin.com/), or even the code cells in [ObservableHQ](https://observablehq.com/demo)) to display & run sample code would reduce friction for developers experimenting with existing GL JS examples. The overhead of copy-pasting entire HTML files to tweak a single line of JavaScript is quite high -- and is not representative of how easy it is to customize and build complex Mapbox-based applications off of existing examples.  

An HTML-only IDE would be a great first step (By "HTML-only", I mean an IDE that only supports editing a single HTML file. CSS and JS would remain inline in the form of `<style>` and `<script>` tags). In the long run, it would be useful to modularize the examples (i.e. use external `.js` script files for JavaScript instead of `<script>` tags), since the boilerplate HTML used by each is virtually identical. 

Additionally, importing GeoJSON data as a variable rather than defining the variable within the HTML's `<script>` tags would help de-clutter examples. For example, consider the `places` variable in the [Filter symbols by toggling a list](https://docs.mapbox.com/mapbox-gl-js/example/filter-markers/) code snippet. This `FeatureCollection` could live in a separate `.json` file to make the code more readable. To maintain the benefit of brevity that comes from keeping all the code in a single file, perhaps an option to toggle between the `modular` and `single file` options could be added.

cc @mapbox/docs

katydecorah:
@adrianababakanian I think this is excellent! I started looking into options to make example page code interactive:


- CodePen has [prefill embeds](https://blog.codepen.io/documentation/prefill-embeds/), which will take code blocks on a given page and turn it into an interactive environment.
- [JSFiddle has embeds](https://docs.jsfiddle.net/embedding-fiddles), but it appears we would have to already create the fiddle in order to embed it.
- [CodeMirror](https://codemirror.net/) appears to be the standard in powering JavaScript web text editors (like CodePen, JSFiddle). If we needed to build it custom, I imagine we’d start there.

(I'm also happy to hear other suggestions, there are the ones that I had top of mind.)

[...]

PR #147, opened 2019-11-20
Add Edit in JSFiddle and CodePen buttons to each example

katydecorah:
This PR adds buttons that will import a given example page code into JSFiddle or CodePen.

* Code is prepped and formatted so that it renders well in either service.
* Sends metadata (title, description, tags) and a link back to docs.mapbox.com along with code.
* Adds a Segment event so we can track how often either button is clicked.

> image

_New edit buttons appear at the top of the example code_

## CodePen

image
_Code imported into CodePen_

image
_Metadata such as title, description, tags, and a link back to the original example are included_

## JSFiddle
image
_Code imported into JSFiddle_

image
_Metadata such as title, description, and a link back to the original example are included (JSFiddle does not support markdown, so it's stripped)_

## For the reviewer:
* You can test out any example by running this site locally.
* While testing locally and logged in, the site will provide a staging token in the example code. When you import to CodePen or JSFiddle that staging token won't work, but you can paste in a production token to verify.

Refs https://github.com/mapbox/mapbox-gl-js-docs/issues/139


docs/components/example.js
```diff
@@ -126,7 +137,15 @@ ${html}
                                 className="bg-white"
                                 data-swiftype-index="false"
                             >
-                                <div id="code">
+                                <div id="code" className="relative">
+                                    <EditButtons
```
adrianababakanian: 💯

docs/components/edit-buttons.js
```diff
@@ -0,0 +1,131 @@
...
+class EditButtons extends React.Component {
+    render() {
+        let { css, code, frontMatter, rawHtml, head, url } = this.props;
+        // regex to find and extract css and scripts in html
+        const srcRegex = /(?<=\bsrc=["'])[^"']*/g,
+            hrefRegex = /(?<=\bhref=["'])[^"']*/g,
+            scriptRegex = /<script>((.|\n)*)<\/script>/,
+            cssRegex = /<style>((.|\n)*)<\/style>/,
+            moreCss = rawHtml.match(cssRegex);
+        // output for code panels
```
adrianababakanian: Really big fan of how you've handled the JSFiddle/CodePen handoff to take of advantage of the modularized code editing that these IDEs provide. Definitely a huge improvement over the current HTML copypasta flow that we've got going on 😄

PR #149, opened 2019-11-22
Fix incorrect closing tag

katydecorah:
The example in this diff has a mismatched closing tag


katydecorah:
This will be fixed in #150 instead where we're adding Prettier and it caught this mismatched tag

PR #155, opened 2019-11-24
:memo: add plugin

dayjournal:
- :memo: add plugin


dayjournal:
Hi,
Is it possible to add my plugin to the plugin list?

PR #166, opened 2019-12-05
add a stretchable icon example

ansis:
Add an example that shows how stretchable icons can be added using `addImage`. The stretchable icons have multiple stretch zones. A version of the image with the stretch zones coloured differently is also included.

Screen Shot 2019-12-05 at 6 41 25 PM


ansis:
To run this locally you have to replace the image urls with the local url:

```diff
 var images = {
-    "popup": "https://docs.mapbox.com/mapbox-gl-js/assets/popup.png",
-    "popup-debug": "https://docs.mapbox.com/mapbox-gl-js/assets/popup_debug.png"
+    "popup": "/mapbox-gl-js/assets/assets/popup.png",
+    "popup-debug": "/mapbox-gl-js/assets/assets/popup_debug.png"
 };
```

ryanhamley:
This doesn't look the same when I run it locally

Screen Shot 2019-12-05 at 4 19 53 PM

docs/pages/example/add-image-stretchable.html
```diff
@@ -0,0 +1,145 @@
...
+    container: 'map',
+    style: 'mapbox://styles/mapbox/streets-v11'
+});
+
+var images = {
+    "popup": "https://docs.mapbox.com/mapbox-gl-js/assets/popup.png",
+    "popup-debug": "https://docs.mapbox.com/mapbox-gl-js/assets/popup_debug.png"
+};
+
+console.log(mapboxgl);
```
ryanhamley: console.log

ansis:
You may need to update your submodule

andrewharvey:
I also couldn't work out how to get test it, even after running `git submodule update`, but apart from that looks good to me!

ansis:
It looks like https://github.com/mapbox/mapbox-gl-js-docs/blob/publisher-production/docs/pages/dist links to the `mapbox-gl-js` directory in the same directory this repo is in. So you may need to checkout the right version and `yarn run build-prod-min` there or something like that

docs/pages/example/add-image-stretchable.html
ansis: fixed in f22496687bb59d3aae358ad6e144637ab8136f72

PR #171, opened 2019-12-11
Show full expression support in style properties

asheemmamoowala:
Following up on https://github.com/mapbox/mapbox-gl-js/issues/7207#issuecomment-562867714, I noticed that we already have information in the style-spec to support which expressions are supported in what properties. 

This PR is a work in progress. In trying to add support for feature state, I discovered that interpolate expressions were not being surfaced as the code was still referring to interpolate functions. Post fix the docs would look like:

image

Having fixed that actually uncovers the issue that I'd like to get some feedback on here:
Each property in the style spec has has an expression section of the form:
```(json)
"expression": {
        "interpolated": true,
        "parameters": [
          "zoom",
          "feature",
          "feature-state"
        ]
      },
```
How should this data be surfaced? The previous method would lead to a growing list of `Supports 'x' expression` statements`. 

Some options I considered:
`Supports x, y, z expressions.`
` Supported expressions: x, y, z`
`Expression support: [icon-zoom], [icon-interpolate],...`

cc @mapbox/gl-js


asheemmamoowala:
@katydecorah Did you have thoughts on the questions at the bottom of the OP: 

How should this data be surfaced? The previous method would lead to a growing list of Supports 'x' expression statements`.

Some options I considered:
`Supports x, y, z expressions`.
`Supported expressions: x, y, z`
`Expression support: [icon-zoom], [icon-interpolate],..`.

chloekraw:
@asheemmamoowala Overall, I like this. If I'm understanding this correctly, "Supports `format` expressions." would be added to the description for `text-field` as the only property that supports `format`.

For style-spec properties that can be overridden in the `format` expression, would you agree with adding "`format` overriding" (or something similar) as a new row under "data-driven styling" in the SDK compatibility table? That would also enable us to clarify which version this behavior is supported for each property and each SDK.

> How should this data be surfaced? The previous method would lead to a growing list of Supports 'x' expression statements`.

I'd say your first option, `Supports feature-state, interpolate expressions.`, is the syntax that best matches the current voice of the docs. For consistency and clarity, I'd also propose that that list of supported expressions is always populated according to alphabetical order.

cc @alexshalamov

asheemmamoowala:
> "Supports format expressions." would be added to the description for text-field as the only property that supports format.

That is not quite the case here. `formatted` is a type and `text-field` accepts a value of that type, so this PR does not affect that property.

> For style-spec properties that can be overridden in the format expression, would you agree with adding "format overriding" (or something similar) as a new row under "data-driven styling" in the SDK compatibility table? 

This was implemented  in https://github.com/mapbox/mapbox-gl-js/pull/9030 and is live in the style spec docs at https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#types-format

image

asheemmamoowala:
@chloekraw I've combined the expression support sentences into one. What do you think?
 
Screen Shot 2020-02-06 at 11 30 33 PM

@katydecorah I'm new to react - so I've used a naive approach to constructing the `Supports (interpolate | feature-state) expressions.` sentence. It works, but Im happy to change it if this is not the correct way to do it. @danswick suggested leaving it as-is since the whole `item` component needs a refactoring 🤷‍♂

docs/components/style-spec/item.js
```diff
@@ -227,14 +227,30 @@ export default class Item extends React.Component {
...
-                    {this.props.function === 'interpolated' && (
+                    {this.props.expression && 
+                        (this.props.expression.interpolated || 
+                            this.props.expression.parameters.includes('feature-state')) && (
                         <React.Fragment>
                             Supports{' '}
-                            <a href="/mapbox-gl-js/style-spec/expressions/#interpolate">
+                            {this.props.expression.parameters.includes('feature-state') && (
+                                <em className="color-gray">
+                                    <a href="#expressions-feature-state">
```
katydecorah: We'll want to update this path to point to the new expressions page:

```suggestion
                                    <a href="/mapbox-gl-js/style-spec/expressions/#feature-state">
```

chloekraw:
This looks good to me. Noting that in the `One of...` sentences, we already use `,` as the delimiter even if there are only two items in the list. So I would prefer to use `,` for `Supports...` as well, instead of introducing the `and` that we'll have to refactor away in the future.

Screen Shot 2020-02-10 at 5 56 37 PM

PR #173, opened 2019-12-12
Update measure example

### LINKED ISSUE #106: Feedback for /mapbox-gl-js/example/measure/
*This feedback was submitted by a user through the feedback widget on the bottom of the page.*

* Page: [/mapbox-gl-js/example/measure/](https://docs.mapbox.com/mapbox-gl-js/example/measure/)
* Section: null
* Date received: 2019-10-08T13:45:43.401Z
* Was this page helpful?: 😕

Feedback:
> The turf.lineDistance() doesn't exist anymore. Its turf.length().

life777:
Fixes #106

This PR updates Turf library, link to Turf documentation and used method for calculating distance.


danswick:
This PR looks good to me. My only question is whether we have any concerns about moving from a Mapbox-hosted plugin to one hosted on npm. I don't think this is an issue and, either way, this example is using a plugin stored in the old `mapbox.js` plugins repository, which isn't idea. @mapbox/docs - does anyone else have any concerns here? If not, I think we should go ahead and merge this!

asheemmamoowala:
>how do you feel about linking to a version of Turf hosted on npmcdn/unpkg? Would it be worthwhile to update our own hosted plugin instead?

It would be better to host turf ourselves but isn't a  blocker for this PR.

PR #177, opened 2019-12-20
remove source inlineing

andrewharvey:
https://github.com/mapbox/mapbox-gl-js/issues/5728 seems to have strong support for not using source inlining, and it keeps coming up as an issue with new users (there are a number of SO, github tickets relating to this), so I'm proposing to avoid it in the examples.

- especially for beginners, source inlining can lead to an incorrect mental model (thinking layers contain sources, instead of in reality layers point to a source) about layers and sources which can subsequently lead to issues with
 - multiple layers from one source (otherwise one is inlined and it can be unclear)
 - removeLayer doesn't remove the source (inclear when inlined)

cc @jingsam @stevage @pathmapper @jimutt

closes https://github.com/mapbox/mapbox-gl-js/issues/5728


andrewharvey:
This is ready to be merged.

PR #190, opened 2020-01-07
Data join using feature state

jbranigan:
I updated the data join example to use `setFeatureState`. Feature state is easier to reason about, and it's the method we recommend to customers. Hope this helps!

### DIFF
--- a/docs/pages/example/data-join.html
+++ b/docs/pages/example/data-join.html
@@ -64,7 +64,7 @@
     ];
 
     
-    map.on('load', function () {
+    map.on('load', function() {
         // Add source for state polygons hosted on Mapbox, based on US Census Data:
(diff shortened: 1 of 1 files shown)


asheemmamoowala:
@jbranigan This exact change was made previously, but reverted in https://github.com/mapbox/mapbox-gl-js/pull/6960. We should hold this update until the next release which will include https://github.com/mapbox/mapbox-gl-js/pull/8987 and reduce developer confusion. 

cc @sansumbrella

asheemmamoowala:
@jbranigan The 1.7.0 release is alive and ready to be used with the `promoteId` property in this example. 

cc @mourner

planemad:
PR for a separate data join with feature state example using the mapbox countries tileset https://github.com/mapbox/mapbox-gl-js-docs/pull/429

PR #191, opened 2020-01-07
Fix references with '#' characters

### LINKED ISSUE #121: Anchor links incorrectly generated in API documentation
## Steps to reproduce
The anchor links generated for each `a` tag in the [Mapbox GL JS API documentation](https://docs.mapbox.com/mapbox-gl-js/api/) are generated incorrectly. For example, consider the Map#addLayer link under the [CustomLayerInterface](https://docs.mapbox.com/mapbox-gl-js/api/#customlayerinterface) section. The generated link is `https://docs.mapbox.com/mapbox-gl-js/api/#mapaddlayer`, when it should in fact be `https://docs.mapbox.com/mapbox-gl-js/api/#map#addlayer`.

## To fix
Change the generated URLs to `/#map#{methodname}` instead of `/#map{methodname}`.

asheemmamoowala:
Closes #121 and https://github.com/mapbox/mapbox-gl-js/issues/4568

Updated the link resolution to explicitly handle link tags with `#` characters because they are not preserved when passed through the `GithubSlug` module

### DIFF
--- a/docs/components/api-item-member.js
+++ b/docs/components/api-item-member.js
@@ -11,7 +11,6 @@ const linkerStack = new LinkerStack({}).namespaceResolver(docs, namespace => {
     const slugger = new GithubSlugger();
     const names = namespace.split('#');
     const nv = names.map((v) => `#${slugger.slug(v)}`).join('');
-    console.log(`slugger transformed ${namespace} to ${nv}`);
     return nv;
 });
 
--- a/docs/components/api-item-member.js
+++ b/docs/components/api-item-member.js
@@ -9,7 +9,10 @@ import Icon from '@mapbox/mr-ui/icon';
 
 const linkerStack = new LinkerStack({}).namespaceResolver(docs, namespace => {
     const slugger = new GithubSlugger();
-    return `#${slugger.slug(namespace)}`;
+    const names = namespace.split('#');
+    const nv = names.map((v) => `#${slugger.slug(v)}`).join('');
+    console.log(`slugger transformed ${namespace} to ${nv}`);
+    return nv;
 });
 
 const formatters = createFormatters(linkerStack.link);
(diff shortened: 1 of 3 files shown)


katydecorah:
@asheemmamoowala Thanks for fixing this! I identified another adjustment we should consider making.

Looking at: https://docs.mapbox.com/mapbox-gl-js/api/#customlayerinterface this PR will fix `Map#triggerRepaint` but not `Map.event:webglcontextlost`. 



image

GitHubSlugger is removing the needed punctuation, where`#mapeventwebglcontextrestored` needs to be `#map.event:webglcontextlost`. I think we could catch instances of `Map.` with something like:

```js
    if (namespace.indexOf('Map.') > -1) {
        return `#${namespace.toLowerCase()}`;
    } else {
        const slugger = new GithubSlugger();
        const names = namespace.split('#');
        const nv = names.map(v => `#${slugger.slug(v)}`).join('');
        if (namespace.startsWith('Map.')) console.log(namespace, names);
        return nv;
    }
```

Also: should we update linkStacker in `api-item-member.js` as well?

asheemmamoowala:
> Also: should we update linkStacker in api-item-member.js as well?

👍 

> GitHubSlugger is removing the needed punctuation, where#mapeventwebglcontextrestored needs to be #map.event:webglcontextlost. I think we could catch instances of `Map.` with something like:

I find that here the `GithubSlugger` is not serving our needs. It doesn't look like we are trying to find unique slugs, but just generate slugs for our needs. A quick search found that the [slugger](https://github.com/HenrikJoreteg/slugger) library offers the flexibility to do what we want out of the box without having to build this up ourselves.

We could switch to this library or write our own `slug` module for this project.

katydecorah:
@asheemmamoowala I think the problem is that we need to slugify the `id` with GitHubSlugger:

https://github.com/mapbox/mapbox-gl-js-docs/blob/12ad32ab3fd71dbeb57ee17212058128bc2df5ee/docs/components/api-item-member.js#L33

From what I can tell, the description gets pass through the formatter, where all links are getting slugified, but in some occasions, such as with `webcontextlost`, the slugified link doesn't match the unformatted `member.namespace`.

https://github.com/mapbox/mapbox-gl-js-docs/blob/12ad32ab3fd71dbeb57ee17212058128bc2df5ee/docs/components/api-item.js#L53

PR #193, opened 2020-01-10
Add information about referrer policies

kkaefer:
This PR adds information about referrer policies and required headers when Mapbox GL JS is used with Mapbox API calls and a [URL-restricted access token](https://docs.mapbox.com/accounts/overview/tokens/#url-restrictions).


ErinQuinn:
JFYI @mapbox/identity-and-auth - @kkaefer is updating the GL JS documentation with advice for developers using an access token with URL restrictions. Looks 👍 to me. Thanks @kkaefer!

PR #218, opened 2020-02-05
Improve how *-transition properties are documented

### LINKED ISSUE #217: Document "*-transition" properties
_From @jfirebaugh on September 22, 2014 18:19_

Document how `*-transition` properties work.

_Copied from original issue: mapbox/mapbox-gl-style-spec#171_

asheemmamoowala:
Closes #217 by describing individual transition properties and how to find them in the style specification.

Before: 
image

After: 
image


docs/pages/style-spec/transition.md
```diff
@@ -17,7 +18,18 @@ A `transition` property controls timing for the interpolation between a transiti
     2
 )}}
 ```
+Any transitionable style property, marked by {{<Icon
+        name="opacity"
+        inline={true}
+    />}}, may also haves its own `*-transition` property that defines specific transition timing for that specific layer property, overriding the global `transition` values.
```
katydecorah: Small typo:

```suggestion
    />}}, may also have its own `*-transition` property that defines specific transition timing for that specific layer property, overriding the global `transition` values.
```

katydecorah: This sentence also has the word "specific" twice. If you feel both instances are important to conveying the meaning, I think you could leave it. Otherwise, I would recommend removing the second instance of "specific".

> defines specific transition timing for that layer property

PR #222, opened 2020-02-16
Update fallback-image.js description 

danswick:
Per some user feedback, it is not immediately clear how this example works. This PR updates the description to make it clear that the fallback magic is in the coalesce expression.


docs/pages/example/fallback-image.js
```diff
@@ -1,6 +1,6 @@
 /*---
 title: Use a fallback image
-description: Fallback to another image when a requested image is not available.
+description: Use a [coalesce](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#coalesce) expression to fall back to another image when a requested image is not available.
```
colleenmcginnis: I wonder if part of the confusion is unfamiliarity with the term "fallback" -- what do you think of using a more generic word/phrase to describe what is happening?

```suggestion
description: Use a [coalesce](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#coalesce) expression to display another image when a requested image is not available.
```

katydecorah:
@danswick ok to merge?

Issue #223, opened 2020-02-17
Adjust zoom level on stretchable icons example


colleenmcginnis:
From @segheysens:

In the [Stretchable Icons example](https://docs.mapbox.com/mapbox-gl-js/example/add-image-stretchable/), the demo map is initialized at too low of a zoom level to see the data layer. If you zoom out slightly, there you can see the icons, but you can't initially.

![stretchable-icons](https://user-images.githubusercontent.com/10479155/74672370-fad28680-5161-11ea-9e73-061925f0a09b.gif)

cc @mapbox/docs

PR #224, opened 2020-02-18
Add default zoom to strechable image example

adrianababakanian:
Closes https://github.com/mapbox/mapbox-gl-js-docs/issues/223


docs/pages/example/add-image-stretchable.html
```diff
@@ -3,7 +3,8 @@
 <script>
     var map = new mapboxgl.Map({
         container: 'map',
-        style: 'mapbox://styles/mapbox/streets-v11'
+        style: 'mapbox://styles/mapbox/streets-v11',
+        zoom: 2
```
ansis: Thanks! Can this be made smaller so that the fifth icon is visible at the bottom? `0` would be ideal but I think https://github.com/mapbox/mapbox-gl-js/issues/7666 prevents us from doing that so maybe `0.1`?

adrianababakanian: Sure thing @ansis , thanks for the review!

PR #226, opened 2020-02-19
Center the map on drone location in live-geojson

adrianababakanian:
Previously, this example appeared to be broken, since the drone icon's location was mostly outside of the map's fixed viewport.

This PR parses the GeoJSON from https://wanderdrone.appspot.com/ and passes it to the `flyTo` method to ensure that the drone icon is always visible on the map. 

That being said, the end visual result is a bit jumpier than ideal. I considered adding logic to only change the map's viewport once the drone icon reached a certain threshold from the currently rendered boundaries, but ultimately decided that might be more overhead than necessary for the functionality this example is trying to get across.


adrianababakanian:
I realized per https://github.com/mapbox/mapbox-gl-js/blob/0412fdb247f0f0c0bdb46d1e1465a848e1eea7dc/src/ui/camera.js#L911-L915 that the `essential: true` option was contributing to the jumpiness, since it causes a fall-through from `flyTo` to `jumpTo`. Removing makes the end result a bit smoother.

andrewharvey:
> I realized per https://github.com/mapbox/mapbox-gl-js/blob/0412fdb247f0f0c0bdb46d1e1465a848e1eea7dc/src/ui/camera.js#L911-L915 that the essential: true option was contributing to the jumpiness, since it causes a fall-through from flyTo to jumpTo. Removing makes the end result a bit smoother.

hmm, the behaviour in your PR looks fine to me (with prefersReducedMotion off) and this animation isn't essential so no need to have essential: true, so looks good to me.

andrewharvey:
I guess now there are two requests made (one internal to setData and one in the example code) which isn't ideal, and it does complicate the example a bit more, which may make it harder to understand.

I guess some people will want this use where the viewport follows the data, but plenty of other use cases won't.

In my opinion https://github.com/mapbox/mapbox-gl-js/issues/6658 is the root cause of the issue here since the request to set the initial zoom at 0 is ignored due to the bug in gl-js.

A possible workaround is to just set the zoom to 0.1 which is zoomed out enough to always see the icon.

arindam1993:
👋 
Imo an example should showcase best practice, and making two network requests for the same data is no bueno.

Could you move make the xhr call first, and then use the downloaded json to set center and `setData` ?

PR #230, opened 2020-02-22
Add canvas source example

malwoodsantoro:
I was recently working on a project that required a canvas source and realized we don't have any examples that show how to use it. It took me quite a bit of time to get it right - so hoping this will help create a starting point for others in the same situation.


colleenmcginnis:
Looks like a good addition to me, @malwoodsantoro! I'm going to defer to @ryanhamley to review since I don't have experience working with `canvas`.

docs/pages/example/canvas-source.html
```diff
@@ -0,0 +1,92 @@
+<canvas id="canvasID" width="400" height="400"> Canvas not supported</canvas>
```
ryanhamley: Extra space before the text

docs/pages/example/canvas-source.html
```diff
@@ -0,0 +1,92 @@
...
+        map.addSource('canvas-source', {
+            type: 'canvas',
+            canvas: 'canvasID',
+            coordinates: [
+                [91.4461, 21.5006],
+                [100.3541, 21.5006],
+                [100.3541, 13.9706],
+                [91.4461, 13.9706]
+            ],
+            animate: true
```
ryanhamley: I think it might be good to add a brief comment here noting when you'd use `animate: true` vs `animate: false`

PR #234, opened 2020-02-22
make the geocoder graphically outside the map in the geocoder outside the map element example

### LINKED ISSUE #39: Example of geocoder control placed outside the map...isn't.
The geocoder control is, um, not outside the map.

screenshot 2019-02-05 17 15 54

I'm not sure exactly what's going wrong here. I suspect maybe because the map is taking up 100% of the container it's within, there's just nowhere for the control to be that isn't in top of the map.

The simplest fix is to change the styling of `#map` to be `top: 100px;` instead of `top: 0;`

andrewharvey:
#39 describes two reports from users who found the geocoder outside the map example because although this example is there to show the geocoder external to the map element in the DOM, such that you can place it anywhere, by still having it graphically on top of the map in the example it led to confusion.

This change aims to avoid such confusion in the future by making the geocoder graphically outside the map in addition the geocoder element to it being outside the map element already.

Screenshot from 2020-02-23 00-24-01

cc @stevage

fixes #39


docs/pages/example/mapbox-gl-geocoder-outside-the-map.html
```diff
@@ -14,11 +14,13 @@
         width: 50%;
         left: 50%;
         margin-left: -25%;
-        top: 20px;
```
katydecorah: @andrewharvey maybe keep the `top` but set to `10px`? This will prevent the shadow from getting clipped:

image

andrewharvey: good point, I've changed that now.

katydecorah:
@andrewharvey if you update this branch with publisher-production the tests should pass

PR #240, opened 2020-02-26
add type cases for layout and paint

katydecorah:
If you visit https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/#layout or https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/#paint and click on the link at "Optional layout" or "Optional paint" it will send you to the types page. I checked out the code before we split Style Spec into pages and confirmed that this link was also pointing at the types page:

image

This PR adds a case for `layout` and `paint` to link to the #layout-property and #paint-property sections.


docs/components/style-spec/item.js
```diff
@@ -82,6 +82,24 @@ export default class Item extends React.Component {
...
+                );
+            case 'paint':
+                return (
+                    <span>
+                        {' '}
+                        <a href="/mapbox-gl-js/style-spec/layers/#paint-property">
+                            paint
+                        </a>
+                    </span>
+                );
```
katydecorah: For some reason prettier formatted a bunch of code on this page (re: lines 248-284), but these lines are the only section that has meaningful changes in this branch.

PR #241, opened 2020-02-27
Fix Style Spec redirects for hashes that begin with `#paint` and `#layout`

katydecorah:
We received feedback that https://docs.mapbox.com/mapbox-gl-js/style-spec/#paint-fill-fill-color is not linking the user to correct page (should be: https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/#paint-fill-fill-color). While testing, I also noticed that hashes that begin with `#layout-` need a slight adjustment as well.

This PR:

- Adds a redirect case for sending hashes that start with `#paint-*` to the layers page and preserve the hash location.
- Fixes redirects for hashes that start with `#layout-*` to preserve the original hash and still send to the layers page.


danswick:
@katydecorah we got some similar feedback this morning for `Map#on` links. I wonder if those are caused by a similar issue https://sentry.io/organizations/mapboxcom/issues/1543127641/?project=1848287&referrer=slack.

katydecorah:
@danswick I think the issue the user reported is related to: https://github.com/mapbox/mapbox-gl-js-docs/issues/121

PR #248, opened 2020-03-11
Add an example for padding

arindam1993:
This adds an example for the new 
https://github.com/mapbox/mapbox-gl-js/pull/8638 
asymmetric viewport feature. This should be merged only after release 1.9.0 is out since that contains new features that are required for this example to run.

A demo of this example is running here.
https://jsfiddle.net/hat4bfod/

I'm concerned this example has a LOT of css, please lmk what you think.


docs/pages/example/add-floating-sidebars-with-padding.html
```diff
@@ -0,0 +1,237 @@
...
+        color: #0aa1cf;
+        cursor: pointer;
+    }
+
+    .sidebar {
+        transition: transform 1s;
+    }
+
+    /*
+  The sidebar stuling has them "expanded" by default, we use CSS transforms to push them offscreen
```
ryanhamley: typo: `stuling`, also this should be two sentences, not a comma

asheemmamoowala: What is stuling? I haven't come across this term before.

docs/pages/example/add-floating-sidebars-with-padding.html
```diff
@@ -0,0 +1,237 @@
+<script src="https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-geocoder/v4.4.2/mapbox-gl-geocoder.min.js"></script>
+<link
+    rel="stylesheet"
+    href="https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-geocoder/v4.4.2/mapbox-gl-geocoder.css"
+    type="text/css"
+/>
+<!-- Promise polyfill script required to use Mapbox GL Geocoder in IE 11 -->
+<script src="https://cdn.jsdelivr.net/npm/es6-promise@4/dist/es6-promise.min.js"></script>
+<script src="https://cdn.jsdelivr.net/npm/es6-promise@4/dist/es6-promise.auto.min.js"></script>
```
asheemmamoowala: Remove the geocoder control. It doesn't add to the understanding of the `padding` options.

docs/pages/example/add-floating-sidebars-with-padding.js
```diff
@@ -0,0 +1,13 @@
+/*---
+title: Add expandable floating sidebars using padding
```
ryanhamley: The title seems to focus more on the "expandable floating sidebars" rather than just having an element overlaying the map. But I think I agree with @asheemmamoowala that we should simplify this to just be a static sidebar or a toolbar or something.

docs/pages/example/add-floating-sidebars-with-padding.js
```diff
@@ -0,0 +1,13 @@
...
+description: >-
+  Offset the [vanishing point](https://en.wikipedia.org/wiki/Vanishing_point) of the map using padding, when expanding floating sidebars.
```
ryanhamley: Would this be easier for people to understand if we just talked about moving the map's center with padding?

docs/pages/example/add-floating-sidebars-with-padding.html
arindam1993: My intention with adding it was to demo how expanding sidebars can keep the pin locked to the cneter as the sidebar expands and collapses.
It also illustrates how to adjust the css so controls shift along with the sidebars.

docs/pages/example/offset-vanishing-point-with-padding.html
```diff
@@ -0,0 +1,151 @@
...
+        color: #0aa1cf;
+        cursor: pointer;
+    }
+
+    .sidebar {
+        transition: transform 1s;
+        z-index: 1;
+    }
+
+    /*
```
asheemmamoowala, line 65: nit: whitespace

docs/pages/example/offset-vanishing-point-with-padding.html
```diff
@@ -0,0 +1,151 @@
...
+
+    function toggleSidebar(id) {
+        var elem = document.getElementById(id);
+        var classes = elem.className.split(' ');
+        var collapsed = classes.indexOf('collapsed') !== -1;
+
+        var padding = {};
+
+        if (collapsed) {
+            // Remove the 'collapsed' class from the class list of the element, this sets it to default state which is expanded.
```
asheemmamoowala, line 127:
```suggestion
            // Remove the 'collapsed' class from the class list of the element, this sets it back to the expanded state.
```

docs/pages/example/offset-vanishing-point-with-padding.html
```diff
@@ -0,0 +1,151 @@
...
+            classes.splice(classes.indexOf('collapsed'), 1);
+
+            padding[id] = 300;
```
asheemmamoowala, line 130: nit: Add a note to indicate that this number matches the width of the sidebar(s)

docs/pages/example/offset-vanishing-point-with-padding.js
```diff
@@ -0,0 +1,12 @@
+/*---
+title: Offset the vanishing point using padding
+description: >-
+  Reduce distortion when using large floating sidebars by offseting the [vanishing point](https://en.wikipedia.org/wiki/Vanishing_point) of map based on sidebar size.
```
asheemmamoowala, line 4:
```suggestion
  Offset the apparent center or [vanishing point]((https://en.wikipedia.org/wiki/Vanishing_point) of the map to reduce distortion when floating sidebars are displayed over the map. 
```

ryanhamley: `sidebars` is too specific. what about `when elements are displayed over the map`?

Issue #250, opened 2020-03-16
[Proposal] Break Mapbox GL JS API Reference into multiple pages


katydecorah:
The documentation team would like to propose breaking the [Mapbox GL JS API Reference](https://docs.mapbox.com/mapbox-gl-js/api/) into multiple pages. We recently completed this work with the [Style Specification](https://docs.mapbox.com/mapbox-gl-js/style-spec/) page.

By breaking the API Reference into multiple pages:

- 👍 Users will be able to find content easier through docs search and other search engines.
- 👍 We will be able to structure content to make it less overwhelming and more palatable to new users.
- 👍 Users will experience faster page loads.
- 👍 The sidebar will be easier to digest and navigate for users.
    - The sidebar will use [dr-ui](https://github.com/mapbox/dr-ui)’s SectionedNavigation component which is used by all documentation pages to help deliver a consistent experience for our users. 
    - We recently [rolled out improvements to this component](https://github.com/mapbox/mapbox-gl-js-docs/pull/245) so that the sidebar will scroll to the active item on load, which helps keep the active item in view for the user so they don’t lose their spot. You can see it in action on the [Style Specification](https://docs.mapbox.com/mapbox-gl-js/style-spec/) page.
- 👍 Our team will have improved data around usage. 
    - Each week the API Reference receives approximately 50,000 page views. Due to the page’s length, it’s difficult for us to determine what content is considered popular with users.
    - After breaking the Style Specification into pages we learned the the Layers page is the most popular Style Spec page.
- 👎 Users can no longer command+F on the entire API Reference since the content will be organized across pages. However, since the page has grown so long, we’ve had reports that command+F can fail to function on this page.

## Mock ups

The follow screenshots are examples of what the proposed pages and sidebar hierarchy could look like.

> ![image](https://user-images.githubusercontent.com/2180540/77317643-f578d680-6ce1-11ea-91f4-090bfe9708fb.png)
> Example of the Home page of API Reference

> ![image](https://user-images.githubusercontent.com/2180540/77317766-2fe27380-6ce2-11ea-8a67-ac05e5965459.png)
> Example of the Map page with all its properties organized in the sidebar. When a sidebar item is scrolled to or clicked, the item’s subordinate hierarchy appears. This interaction will be the same as how a page like [Layers](https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/) reveals subordinate hierarchy on the same page.

[...]

PR #260, opened 2020-03-25
[api-pages] Create structure for API Reference markdown pages, use NavigationAccordion 

katydecorah:
**This PR merges into `api-ref-main`.** (We will create a PR for `api-ref-main` once it has commits as it will be our main branch for this work).

This PR provide the structure for the work to come in breaking the API Reference into pages (ref #250):
- Adds markdown pages for the API Reference (content forthcoming).
- Updates the PageShell so that the API Reference uses NavigationAccordion.
- Moves the page's dataset for the navigation to the `data/` folder, this is the same pattern we use  for all other pages that use NavigationAccordion.


danswick:
This is great! I'm sure we'll need to tweak things once we get some momentum, but this looks like a great structure to me.

PR #261, opened 2020-03-25
[api-pages] create `api` folder in components; move over and rename api components

katydecorah:
**This PR merges into api-ref-main.**

This PR adds `docs/components/api` directory for us to store current api-related components and future ones.


danswick:
Makes perfect sense to me and matches the style-spec component org 👍

PR #262, opened 2020-03-26
Add within expression metadata

ryanhamley:
Adds metadata for `within` expression so that it shows up in the docs

Screen Shot 2020-03-26 at 4 56 07 PM


docs/components/expression-metadata.js
```diff
@@ -230,6 +230,12 @@ const types = {
             parameters: ['previously bound variable name']
         }
     ],
+    within: [
+        {
+            type: 'boolean',
+            parameters: ['geometries (GeoJSON Feature or FeatureCollection)']
```
chloekraw: @zmiao said on a gdoc comment that GeoJSON Feature and FeatureCollection objects are also accepted. If that’s true, we should add that in here

chloekraw: I meant add "Polygon" and "Multipolygon" here as types that are also accepted

zmiao: `within` supports the 3 following formats:
```
{'type': 'Polygon/MultiPolygon', 'coordinates':[[[...]]]}
 
{
   'type': 'Feature',
   'properties': {},
   'geometry': {'type':'Polygon/MultiPolygon', 'coordinates':[[[...]]]}
}
 
{
  'type': 'FeatureCollection',
  'features': [
    {
      'type': 'Feature',
      'properties': {},
      'geometry': {'type':'Polygon/MultiPolygon', 'coordinates':[[[...]]]}
    }
   ]
}
```

ryanhamley: @zmiao @chloekraw Does this look better now?

docs/components/expression-metadata.js
```diff
@@ -230,6 +230,14 @@ const types = {
             parameters: ['previously bound variable name']
         }
     ],
+    within: [
+        {
+            type: 'boolean',
+            parameters: [
+                'geometries: GeoJSON Geometry | GeoJSON Feature | GeoJSON FeatureCollection (Geometry object(s) must be of type Polygon or MultiPolygon)'
```
chloekraw:
```suggestion
                'geometry, geometry' 
```

docs/components/expression-metadata.js
chloekraw: @zmiao thanks for writing it out!

chloekraw:
Now that I look at existing entries as a reference, here's where I'm at:

- We list types in their singular form.
- The parameter accepted in the within expression is of type `geometry` (ref: [`geometry-type`](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#geometry-type)).
- Other times we've used `a | b | c` have been to distinguish between different types, e.g. in [`length`](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#length).
- We typically don't use the metadata to specify valid inputs within a type. Instead, restrictions on what is supported are described in the description, e.g. see [`match`](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#match), [`>`](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#%3E). 

That led me to comment the suggestion above. Then, I would edit the style spec description to strike the second sentence, add paragraph breaks for clarity, and add the following to the SDK compatibility table:

| SDK Support | GL JS | Android | iOS | macOS |
| ------------- | ------ | -------- | ---- | ------- |
| basic functionality | >= 1.9.0 | >= 9.1.0 | >= 5.8.0 | >= 0.15.0 |
| Polygon | >= 1.9.0 | >= 9.1.0 | >= 5.8.0 | >= 0.15.0 |
| MultiPolygon | >= 1.9.0 | >= 9.1.0 | >= 5.8.0 | >= 0.15.0 |

cc @mourner

ryanhamley:
@chloekraw The `geometry-type` you linked to isn't a type in our system though. It just returns the type of a geometry. I don't think we have an explicit `geometry` type. In any case, it wouldn't strictly be true that the type is `geometry` because the type can be `geometry`, a `feature` or a collection of features (`FeatureCollection`).

chloekraw:
@ryanhamley I agree it's tricky when the type isn't already defined in our system. Do you have an alternate proposal? 

I'm fine with including `geometry | feature` in the docs if everyone else is. I chose to err on the conservative side by only making up one type that has some existing reference in our style spec docs (`geometry`) as opposed to making up two (`geometry` and `feature`), but I can see how omitting it may mislead developers into thinking that `Feature`/`FeatureCollection` is not supported.

On the other hand, I'm comfortable moving forward with my current proposal. We (will) have usage examples and technical guides to help developers understand how to use the `within` expression, and those will make the nuances clear; we can't explain everything in the docs alone.

I'm not wedded to any one path though and am open to alternate suggestions.

zmiao:
@chloekraw @ryanhamley I am taking the reference from [geojson.org](https://geojson.org/),

It illustrates that 

> GeoJSON supports the following geometry types: Point, LineString, Polygon, MultiPoint, MultiLineString, and MultiPolygon. Geometric objects with additional properties are Feature objects. Sets of features are contained by FeatureCollection objects. 

So I wonder shall we use the similar term,  **Geometric objects**, **Feature objects**, **FeatureCollection objects**?

chloekraw:
@zmiao sounds good. I've updated the suggestion to use `object` per latest team discussions. We've also agreed to expand the SDK compatibility table section with the specific supported inputs in the future if/when we add support for other geometries.

ryanhamley:
Screen Shot 2020-04-02 at 4 20 16 PM

PR #267, opened 2020-03-31
[api-pages] Create smaller components from api/item

katydecorah:
**Once https://github.com/mapbox/mapbox-gl-js-docs/pull/261 merges, this PR will merge into api-ref-main.**

This PR: 
* breaks up api/item.js into smaller components to help improve codebase readability
* replaces empty `div` with `React.Fragment` where possible to reduce node size
* removes the Feedback module temporarily as we will be assessing its treatment in another PR (for example, whether to have one Feedback module per page or continue with many per page, ref #269).

Aside from removing the Feedback module temporarily, this PR does not make any visual changes.


docs/components/api/item.js
```diff
@@ -65,347 +56,85 @@ class ApiItem extends React.Component {
...
-                            }}
-                        />
+                        <ClassName formatters={formatters} section={section} />
                     )}
 
-                {section.version && <div>Version: {section.version}</div>}
-                {section.license && <div>License: {section.license}</div>}
-                {section.author && <div>Author: {section.author}</div>}
-                {section.copyright && <div>Copyright: {section.copyright}</div>}
-                {section.since && <div>Since: {section.since}</div>}
```
katydecorah: I don't think this is used at all.

katydecorah:
@danswick this is ready for review!

Since this PR creates a lot of smaller components, I created a temporary commit d56c715 to add back the full page API reference. I figured it would be helpful to be able to do a visual review since nothing visual should change from this PR.

Before we merge this into api-ref-main, we can revert d56c715 or we could keep it until we have a way to shovel content into pages! I'm cool with either.

docs/pages/api/index.js
```diff
@@ -0,0 +1,142 @@
...
+            type: 'root',
+            children: ast.children[0].children.concat(ast.children.slice(1))
+        };
+    }
+    return (
+        <span dangerouslySetInnerHTML={{ __html: formatters.markdown(ast) }} /> // eslint-disable-line
+    );
+}
+
+class Note extends React.Component {
```
danswick, line 45: What do you think about either moving the components in this page to separate files or replacing with dr-ui components?

katydecorah: Sorry! You won't need to review this page, this is the original page and I added back so you can do a visual compare. Ref https://github.com/mapbox/mapbox-gl-js-docs/pull/267#issuecomment-608042691

We'll need to find a new way to slot content into the markdown pages.

PR #268, opened 2020-03-31
[api-pages] Create formatters and empty utilities, add Jest

katydecorah:
**Once #267 merges, this PR will merge into api-ref-main.**

This PR:
* creates util/formatters and replaces duplicated code to point at this single function.
* creates util/empty and replaces duplicated code to point at this single function

This PR does not introduce any visual differences.


katydecorah:
> what do you think about adding a couple of small tests for the util functions?

Good thinking! Added ✅

PR #278, opened 2020-04-16
[docsprint] Add an example for AnimationOptions

ryanhamley:
Screen Shot 2020-04-17 at 2 17 37 PM

This PR adds an example for using `AnimationOptions` to control camera animations


docs/pages/example/camera-animation.html
```diff
@@ -0,0 +1,263 @@
...
+                easing: easingFn,
+                offset: [offsetX, offsetY],
+                animate: animate,
+                essential: true // animation will happen even if user has `prefers-reduced-motion` setting on
+            };
+
+            // Create a random location to fly to by offsetting the map's
+            // initial center point by up to 5 degrees.
+            var center = [
+                -95 + (Math.random() - 0.5) * 10,
```
danswick, line 238: What do you think about increasing the range here? Sometimes the movement is pretty small and it's difficult to perceive what has changed.

ryanhamley:
@danswick Thanks for the feedback. I think I addressed everything

PR #282, opened 2020-04-22
Create new reverse-geocoder example

ZiZasaurus:
While we have examples that allow our users to search with coordinates and others that allow users to search with addresses, we don't have one example that allows users to search using coordinates or addresses. This was requested by a customer previously and I thought it would be great to have.


andrewharvey:
> While we have examples that allow our users to search with coordinates and others that allow users to search with addresses, we don't have one example that allows users to search using coordinates or addresses. This was requested by a customer previously and I thought it would be great to have.

Sorry maybe I'm misinderstanding but isn't that exactly what https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder-accept-coordinates/ is? It allows users to search for an address or via directly entering coordinates?

I think we need to be clear of the distinction between forward geocoding coordinates (in which case the exact coordinates entered are used) and reverse geocoding (where user is presented with places nearby their coordinates and they choose from one of them, but not the exact coordinates entered)

ZiZasaurus:
Hi @andrewharvey, while that example allows users to input coordinates in the search field,  I believe customers are expecting that the coordinates yield actual nearby addresses as returned results, allowing them to select the correct address before the marker is added, rather than just adding a marker at the lat and long they specified.

andrewharvey:
> Hi @andrewharvey, while that example allows users to input coordinates in the search field, I believe customers are expecting that the coordinates yield actual nearby addresses as returned results, allowing them to select the correct address before the marker is added, rather than just adding a marker at the lat and long they specified.

Okay so it's the latter option, true reverse geocoding (whereas the current example is a forward search with coordinates).

In that case it should just be setting `reverseGeocoder: true` https://github.com/mapbox/mapbox-gl-geocoder/blob/master/API.md and not needing to parse out coordinates.

ZiZasaurus:
You're right, I was working off of the original example and completely missed the options! Made the changes. :)

docs/pages/example/mapbox-gl-geocoder-reverse-geocoder.html
```diff
@@ -0,0 +1,22 @@
...
+        zoom: 13
+    });
+
+    var geocoder = new MapboxGeocoder({
+        accessToken: mapboxgl.accessToken,
+        mapboxgl: mapboxgl,
+        localGeocoder: coordinatesGeocoder,
+        placeholder: 'try: -77.033835, 38.89975 or "740 15th Street, NW"',
+        types: 'address',
+        reverseGeocoder: true
```
andrewharvey, line 18: sorry I wrote the wrong option before :facepalm: , it's `reverseGeocode` per https://github.com/mapbox/mapbox-gl-geocoder/blob/master/API.md

ZiZasaurus: No worries! oddly enough, `reverseGeocoder` works but `reverseGeocode` doesn't.

andrewharvey: I checked and the option should be `reverseGeocode` https://github.com/mapbox/mapbox-gl-geocoder/blob/master/lib/index.js#L427.

Using `reverseGeocoder` wouldn't change how the geocoder behaves.

So to make sure I understand, the aim of the example is to allow searching for addresses or searching for coordinates and when the input is coordinates then use the reverse geocoder to show nearby addresses?

ZiZasaurus: Hmm, yes you're correct about `reverseGeocooder`. And yes, that is the aim of the example.

andrewharvey: There was a bug I just merged into mapbox-gl-geocoder master with proximity, so testing this with trackProximity: false as a workaround until we can do a new release of mapbox-gl-geocoder.

If I specify `reverseGeocode: true` then for a normal place search it will still return normal search results, but in the special case of entering coordinates, it will return the nearby place returned by the reverse geocoder, which is what this example is designed for right?

mapbox-gl-geocoder is not designed out of the box to handle coordinates with unless you set reverseGeocode: true or treat them with localGeocoder like the other example.

docs/pages/example/mapbox-gl-geocoder-reverse-geocoder.js
```diff
@@ -0,0 +1,11 @@
+/*---
+title: Search using Coordinates or an Address
```
colleenmcginnis, line 2: We use sentence case for example titles.

```suggestion
title: Search using coordinates or an address
```

docs/pages/example/mapbox-gl-geocoder-reverse-geocoder.html
```diff
@@ -0,0 +1,21 @@
+<div id="map"></div>
+<div id="geocoder" class="geocoder"></div>
```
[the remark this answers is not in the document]
ZiZasaurus: makes sense!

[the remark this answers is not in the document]
colleenmcginnis: 🎉  You should also remove this line.
```suggestion
```

katydecorah:
@ZiZasaurus We made some changes to the examples file format in #374 - instead of JavaScript files we are now using markdown files. I updated your branch to match the new system. Let me know if you have any questions!

ZiZasaurus:
Thank you so much!! No questions from me :)

katydecorah:
@ZiZasaurus hi again :) I updated this branch with a newly required frontmatter item (ref #377) - No actions on your end!

docs/pages/example/mapbox-gl-geocoder-reverse-geocoder.html
```diff
@@ -0,0 +1,29 @@
+<script src="https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-geocoder/v4.6/mapbox-gl-geocoder.min.js"></script>
+<link
+    rel="stylesheet"
+    href="https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-geocoder/v4.6/mapbox-gl-geocoder.css"
+    type="text/css"
```
colleenmcginnis, line 5:
```suggestion
<script src="https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-geocoder/v4.6.0/mapbox-gl-geocoder.min.js"></script>
<link
    rel="stylesheet"
    href="https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-geocoder/v4.6.0/mapbox-gl-geocoder.css"
    type="text/css"
```

PR #286, opened 2020-04-23
Update latest-version.test error message

katydecorah:
While working through open PRs today, I noticed some trouble with out-of-date mapbox-gl-js submodule. Whether it's the test failing because it's out-of-date or the contributor updated the submodule and it caused confusing merge conflicts.

Since the submodule is routinely updated by the gl-js team, I think we should change the error message to instruct the contributor to either update their branch with publisher-production or ask `@mapbox/docs` for assistance.


test/latest-version.test.js
```diff
@@ -25,8 +25,7 @@ test(`Is latest version`, t => {
             res,
             `Mapbox GL JS submodule is not pinned to the latest stable release version.
              Latest: ${tags[0]}. Currently checked out: ${tags[1]}.
-             Check out the most recent with:
-             "git checkout tags/${tags[0]}"`
+             To proceed: update this branch with the latest from publisher-production or notify @mapbox/docs.`
```
adrianababakanian, line 28: This is a bit nit-picky, but as someone who has frequently encountered this message (😅) I think having explicit `git` commands is really useful for guidance. Perhaps something a little more flushed out like this could be nice?
 ```suggestion
             To proceed, pull the latest from publisher-production and merge with "git merge publisher-production" from the feature branch. Or, notify @mapbox/docs.`
```

katydecorah: This is great feedback! Thanks @adrianababakanian

PR #295, opened 2020-05-01
documentationjs configuration

danswick:
Draft PR for determining our path forward to multi-page API docs. 

Our original idea was to use the `documentation.yml` config to give some items more visibility in the nav, but that might not be possible. After a bit of experimentation, I don't think `documentation.yml` will work with namepaths like `Map#getBounds`, but I'm not entirely sure why. [Others have run into the same issue](https://github.com/documentationjs/documentation/issues/768), so I think we are either bumping up against a documentation.js limitation or I'm misunderstanding how documentation.js works (or both).

cc @katydecorah


danswick:
We're going to try another direction.

PR #299, opened 2020-05-04
[api-pages] redirect hash links to new pages

katydecorah:
**This branch merged into `api-ref` branch.**

This PR adds a function to help redirect hash links to its new page. We may need to adjust the mappings and names based on how we finalize the pages, but it's a start.


katydecorah:
> What do you think about combining the two redirect things?

Since these functions work differently, I think we should keep them separate. And also for the reason you listed ;)

Issue #275, opened 2020-04-10
Worldview example modifies line styles incorrectly


planemad:
Toggling worldview sets the layer filters as expected, but also seems to incorrectly modify the existing line styles.  

Possibly a regression, since i recall it used to work fine.

![_Screen](https://user-images.githubusercontent.com/126868/79005669-62370200-7b25-11ea-9d61-ac28477777bb.gif)

Link: https://docs.mapbox.com/mapbox-gl-js/example/toggle-worldviews/

cc @katydecorah

PR #302, opened 2020-05-06
Update toggle worldviews example

colleenmcginnis:
Updates the [Change worldview of administrative boundaries](https://docs.mapbox.com/mapbox-gl-js/example/toggle-worldviews/) example. Changes include:

- Shortening the description and pointing to the `worldview` glossary entry for more details on available worldviews.
- Adding an option for the Japan worldview.
- Narrowing the scope to only include `admin-0-*` just to illustrate how to approach worldviews (but we can add `admin-1-*` if we think it's necessary).
- Adding detailed code comments.
- Fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/275 
- Fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/195 (I think the refactored code in this PR will make it sufficiently easier to adapt for setting worldview on map load)

image

cc @planemad @ClareTrainor @lukasmartinelli


docs/pages/example/toggle-worldviews.js
```diff
@@ -1,11 +1,7 @@
 /*---
 title: Change worldview of administrative boundaries
 description: |
-  Uses the [worldview](https://www.mapbox.com/vector-tiles/mapbox-streets-v8/#-worldview-text) value to adjust administrative boundaries based on the map's audience. You can see the worldview options within the worldviews variable in this example. They are as follows:
-  - **CN**: Boundaries for a mainland Chinese audience/worldview, but not officially approved for use in the PRC.
-  - **IN**: Boundaries conforming to cartographic requirements for use in India.
-  - **US**: Boundaries for an American audience, & which are generally appropriate outside of China & India.
-    Lines do not necessarily reflect official US foreign policy.
+  Uses the `worldview` value to adjust administrative boundaries based on the map's audience. Read more about [worldviews](https://docs.mapbox.com/glossary/worldview/).
```
andrewharvey, line 4:
```suggestion
  Uses the `worldview` value to adjust administrative boundaries based on the map's audience. Read more about [worldviews](https://docs.mapbox.com/help/glossary/worldview/).
```

docs/pages/example/toggle-worldviews.html
```diff
@@ -1,83 +1,149 @@
...
+        var inputs = worldviewOptions.getElementsByTagName('input');
+        for (var i = 0; i < inputs.length; i++) {
+            // After the map has finished loading, enable the
+            // inputs to allow the user to change the worldview.
+            inputs[i].disabled = false;
+            // Specify that the switchWorldview function
+            // (defined below) should run every time an input
+            // is clicked.
+            inputs[i].onclick = switchWorldview;
+        }
```
ryanhamley, line 106: IMO, this adds extra complexity that's not really needed for an example. I had to stop for a second to realize what this was for. My suggestion is to either let the radio buttons always be enabled or else populate the dialog on `load` (maybe with some placeholder text for before the map loads). Either way, you should be able to add the `onclick` function when you create the input.

docs/pages/example/toggle-worldviews.html
```diff
@@ -1,83 +1,149 @@
...
     });
+
+    // Create a function that will identify which worldview
+    // is currently selected and run the filterLayers function
+    // with that worldview as an argument.
+    function switchWorldview(option) {
+        var worldview = option.target.id;
```
ryanhamley, line 113: there's no need to save this to a variable just to pass it to a function one line later, unless you want it just for maximal clarity

docs/pages/example/toggle-worldviews.html
```diff
@@ -1,83 +1,149 @@
...
+        filterLayers(worldview);
+    }
+
+    // Update the filter for each "admin-0-*" layer.
+    // These are the same filters used in the Mapbox Light
+    // style except that the worldview is based on user input.
+    function filterLayers(worldview) {
+        // The "admin-0-boundary-disputed" layer shows boundaries
+        // at this level that are known to be disputed.
```
ryanhamley, line 122: Maybe expand on `this level` and what `admin-0-*` refers to as compared to say `admin-1`

PR #303, opened 2020-05-06
[api-pages] enable scrollspy, use margin-scroll-top; adjust heading hierarchy

katydecorah:

This PR enables scrollspy and fixes heading heirarchy:
- Turned markdown pages in JS pages. Since we use a markdown plugin to sectionize the pages so that scrollspy can determine which section is live, we need to sectionize the pages manually since all the content is generated. 
  - A new component `SectionWrapper` was added to this PR to handle wrapping each section.
  - A new component `ApiItemContents` was added (which is split from `ApiItem`) to enable scrollspy on the Map class page.
  - A new component `Github` was added to show the github link on the Map class page.
- Fixes heading hierarchy for all pages by passing a new `headingLevel` prop.
- Uses [`margin-scroll-top`](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-margin-top) to add spacing to clickable headings to clear the TopBarSticker. (While this doesn't have IE11 support, I think this solution provides the least friction and since it does not prevent users from reading the content - just makes it a little less polished on IE11)


docs/components/api/item-contents.js
```diff
@@ -0,0 +1,158 @@
+import React from 'react';
```
katydecorah, line 1: 💡 this new component was pulled from `docs/components/api/item.js`

docs/components/api/item-member.js
```diff
@@ -16,12 +16,10 @@ class ApiItemMember extends React.Component {
     render() {
         const member = this.props;
         return (
-            <div className="border-b border--gray-light">
-                <div
-                    className="pt60"
-                    style={{ marginTop: '-60px' }}
-                    id={member.namespace.toLowerCase()}
-                />
+            <div
```
danswick, line 19: I see the Div Reduction Act of 2020 is going well!

docs/components/api/section-heading.js
```diff
@@ -1,36 +1,36 @@
...
 export default class SectionHeading extends React.Component {
     render() {
-        const { section } = this.props;
+        const { section, headingLevel } = this.props;
+        const HeadingLevel = `h${headingLevel}`;
         return (
-            <div className="mb24">
-                <h3 className="mb12" id={section.namespace.toLowerCase()}>
+            <React.Fragment>
+                <HeadingLevel
```
danswick, line 11: Interesting! I didn't know you could drop in a string value here like this.

docs/pages/api/events.js
```diff
@@ -1,11 +1,16 @@
----
+/*---
```
danswick, line 1: One thing to consider with md --> js is potentially adding intro text to each page in the future. Though maybe that breaks the pattern too much. 

We could put intro text directly in the `documentation.yml` or in its own `md` file that gets imported into `documentaion.yml` like this example: https://github.com/documentationjs/documentation/blob/master/docs/CONFIG.md#configuring-documentationjs. 

Looking at the site now, it's not obvious that we even need additional intro text or any other content (except, perhaps, the index). 

Let's talk about this!

docs/components/api/item-member.js
```diff
@@ -16,12 +16,10 @@ class ApiItemMember extends React.Component {
...
+                className="border-b border--gray-light scroll-margin-top"
+                id={member.namespace.toLowerCase()}
```
danswick, line 21: I noticed some console errors related to this. I think it's only related to the Map instance members that have two versions (a 2-parameter version and a 3-parameter one that accepts a `layerId'). We should maybe change these in the docs instead of trying to fix here, but I'm not really sure what the best path is.

docs/pages/api/events.js
katydecorah: > We could put intro text directly in the documentation.yml

Per sync: Let's do this! These descriptions were originally in the yml, but I removed them when I thought we were moving towards markdown.

docs/components/api/item-member.js
katydecorah: Per sync, I'll append an index to these keys to make them unique for now. Eventually, we'll see about updating the docs to have a singular `once` and `off`.

PR #304, opened 2020-05-06
[api-pages] move navigation generation as Batfish dataSelector; add test

katydecorah:

This PR:
- moves the generation of the API navigation from the client-side to the server-side. Batfish will now build this data as a dataSelector. (I'd love to do this for style-spec, too, but will ticket later instead.)
- sorts alphabetically to make sure items remain consistently in order.
- adds a test that can help us verify changes to this navigation dataset.


katydecorah:
Going to hold on this one until we decide how we'll handle sorting of methods and events.

PR #308, opened 2020-05-08
[api-pages] reintroduce feedback module

katydecorah:
**This branch merges into `api-ref-main`**

This PR:
- creates a local feedback module to reduce repeated code.
- adds back the Feedback module to the api pages. We are changing the frequency at which the module appears. Following the patterns of our templates, the feedback module will appear after every h2 section, except for the Map page where it will appear once at the end of the page.
- adds a `Section` component in `ApiPageItems` to better differentiate the three layout types.


docs/components/api/api-page-items.js
```diff
@@ -14,23 +15,53 @@ export default class ApiPageItems extends React.Component {
...
+                return <Section key={child.name} {...this.props} {...child} />;
             }
         });
     }
 }
 
 ApiPageItems.propTypes = {
-    pageTitle: PropTypes.string.isRequired
+    pageTitle: PropTypes.string.isRequired,
+    location: PropTypes.object.isRequired
```
danswick, line 42: Oooh yes i meant to ask about `location` last week but forgot!

docs/components/markdown-page-shell.js
```diff
@@ -1,29 +1,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import PageShell from './page_shell';
-import Feedback from '@mapbox/dr-ui/feedback';
-import constants from '../constants';
+import Feedback from './feedback';
 
 class MarkdownPageshell extends React.Component {
-    constructor(props) {
```
danswick, line 8: Love having this in the component now!

katydecorah:
>  I wonder if the density of feedback components is too low on the Map page now.

I hear you. Getting this balance right is difficult. 

I think we should merge this for now and towards the end of the week we can push the main branch to staging and get thoughts from the rest of the docs team. I'll this reminder in our task list doc.

PR #309, opened 2020-05-08
[api-pages] add filter function for members list

katydecorah:
@danswick For pages like Map, the instance member and events in the sidebar are overwhelming. I feel like I'm trying to scroll around them and their utility feels lost. 

I'd like to propose introducing a lightweight filter function for member lists (that will only appear if there are > 5 items). This would allow us to remove the noisy instance members and events from the sidebar, while still allowing the user to navigate this long list.

What do you think?

If you think this is useful, I'd expand the filter function to search within the `descriptions` of each member (it's set to just the member name right now).

2020-05-08 at 5 02 PM
> Demo of filter


docs/components/api/members-list.js
```diff
@@ -4,11 +4,60 @@ import ApiItemMember from './item-member';
...
+            : members;
+
         return (
             <SectionWrapper title={title} {...this.props}>
-                {members.map((member, i) => (
+                {members.length > 5 && (
+                    <div className="relative">
+                        <div className="absolute flex-parent flex-parent--center-cross flex-parent--center-main w36 h36">
+                            <svg className="icon">
+                                <use xlinkHref="#icon-search"></use>
```
colleenmcginnis, line 35: Could consider using the `filter` icon.

katydecorah: I tried this! This icon looks 🤔 not very clear:

image

Thinking we should stick with the spy scope for now. Or we could just remove the icon since it's decorative.

katydecorah:
@colleenmcginnis thanks for the notes!

> More distinction between sections (i.e. instance members and events) with more space or bigger headings might help make it clearer where one filterable list ends and the next begins. (That might only really be an issue with Map class page since the list is sooo long. On the other pages, like Markers and controls, it's much clearer what list is attached to search bar.)


I played with this for a while this morning. I did a couple things:

1. Increased spacing between headings slightly.
2. Introduced new styling of member items by adding a border around the entire item and adding a background to the item heading. This treatment allows us to add back a white background on the item's content, which I think makes the content easier to read.

> I noticed that when you type in the filter bar > click the first result to expand > clear the filter then the first result in the unfiltered list is autoexpanded (there's similar behavior with the second result etc).

This should be fixed now! I set it so that everything will collapse while searching and clearing the search button. 

---

Some other things added:

- Turned each item member heading into an html heading and made sure the heading level cascades properly to it's child elements.
- Fixed spacing on examples.
- Moved the function to handle the filter into it's own file, expanded it's functionality to do a string match in all string values in the member's object, and added tests.

docs/components/api/members-list.js
```diff
@@ -2,14 +2,79 @@ import React from 'react';
...
+        });
+        this.clearHash();
+    };
+
+    // clear hash from window location
+    clearHash = () => {
+        if (history.pushState && window) {
+            history.pushState(null, null, window.location.pathname);
+        }
+    };
```
colleenmcginnis, line 28: I'm not sure if this is new or if it is something you're working on in a different PR, but I'm also seeing warnings in the console related to `location`. Here's an example from the `Map class` page:

image

katydecorah: Thanks! I'll see about fixing in this PR.

katydecorah:
Note for myself in the morning: add aria-expanded to indicate when a member item is open

katydecorah:
@colleenmcginnis I think I've addressed all your comments. Thanks for your 🦅 👁️. This is ready again. 

One change:

- I'd like to try https://fusejs.io/ to power our search. I found the implementation of the library to be much more straightforward than writing our own filter, plus it sorts by score. I feel like we could call the input a search now, rather than a filter. I think this could be a library we use in other places (for example the sidebar search).

PR #310, opened 2020-05-08
Add expressions metadata for distance, slice, index-of

chloekraw:
https://github.com/mapbox/mapbox-gl-js/pull/9655

PR #311, opened 2020-05-09
Add test to assert that every type exists in the style-spec

katydecorah:
This PR adds a test to make sure that every `type` defined in expression-metdata.js exists to help catch this in the future.

#310 updated expression-metadata.js, but the style-spec version doesn't contain `distance` which has caused the publisher build to fail. Since publisher-production is failing, it cannot receive updates. **Let's figure out the next step on Monday to unblock the site.**

## Next actions

@chloekraw @asheemmamoowala I'd love your advice on how to move forward. I see two options so far:

1. Comment out [`distance`](https://github.com/mapbox/mapbox-gl-js-docs/blob/4de6055b6bb8c28960fff2a7d8b4bae902cf5c35/docs/components/expression-metadata.js#L270-L275) from the time being.
2. Update style-spec to a version that includes `distance` - if that exists.

## Note for PR reviewer

I split `expressions.js` from `expression-metadata.js` so we'd get a clean test error message. Otherwise the test suite would error before it could run, which is nice that we'd still receive a failed test, but if we allow the test to run then it will tell us exactly which type is failing.

> image
> Error message from the new test that is pointing at the missing `distance` (which is current failing)

cc @danswick @colleenmcginnis


chloekraw:
@katydecorah thank you for this. Leaving out `distance` from the style spec docs isn't a great option as the expression is released in stable versions on iOS and Android, and we want to be able to talk about it in blogs, emails, etc. without confusing customers. Are there ways to fake this expression's existence in the style spec even though it's not yet implemented in GL JS?

asheemmamoowala:
> Comment out `distance` from the time being.
> Update style-spec to a version that includes distance - if that exists.

@katydecorah I think we should do both of these. The next version of the style-spec will have this expression. See https://github.com/mapbox/mapbox-gl-js/pull/9671.

katydecorah:
@asheemmamoowala sounds like a plan. I'll comment out `distance` in this PR and we'll plan to add it back with the next version.

katydecorah:
I just added one more test that checks that every definition in the style-spec is found in expressions-metadata.js. This will provide a reminder (a failing test) to uncomment `distance` once style-spec is updated.

PR #313, opened 2020-05-10
do something with result in mapless geocoder example

andrewharvey:
If you're using the MapboxGeocoder without a map, then you'd usually want to do something with the result, so we should round out this example and show how to obtain the result.

I'm not sure if there is a policy around console logging in the example, I thought about showing the result somehow on the page, but decided it would add too much overhead to the example, but I don't mind if someone else wanted to add that.


katydecorah:
@andrewharvey How about writing the results to a new element, `<pre id="result"></pre>`, instead?

Something like:

```js
// geocoder results container
var results = document.getElementById('result');

// add geocoder result to container
geocoder.on('result', function (e) {
  results.innerText = JSON.stringify(e.result, null, 2)
});

// clear results container when search is cleared
geocoder.on('clear', function () {
  results.innerText = '';
});
```

katydecorah (approved):
Thank you! I updated to incorporate the idea discussed in https://github.com/mapbox/mapbox-gl-js-docs/pull/313#issuecomment-680909625

PR #315, opened 2020-05-12
[api-pages] Build out API reference introduction and combine with overview

danswick:
I would love some feedback on this PR. It does a couple of things:

1. Removes (sort of) the content from the old intro section from the API reference page. In my opinion, it was too duplicative of the content from the overview page.
2. Moves all of the overview content into the Introduction section. Almost every user lands on the api reference page and we should front-load that page with critical information on how to be successful. I also think it is good to reduce the number of sections nav tabs.
3. Adds a "how to read this documentation" section inspired by https://docs.mapbox.com/api/#reading-this-documentation. 
4. Adds a live map demo to the quickstart section. My hope is that it will be more engaging to have a life example and some enticing map imagery front-and-center. Also updates the copyable quickstart code with `center`, `zoom`, and the latest streets style version 😬. 

Does this shift work? Would it be better to keep the overview section and stop redirecting to the "api reference" page? @colleenmcginnis suggested we may want to turn the "overview" section into more of a "guides" section with more content, but that is probably out of scope for the current docs push. 

image


docs/components/markdown-page-shell.js
```diff
@@ -24,8 +24,7 @@ class MarkdownPageshell extends React.Component {
         return (
             <PageShell {...this.props}>
                 <div className="prose">
-                    {this.props.location.pathname !==
-                        '/mapbox-gl-js/overview/' && (
+                    {this.props.location.pathname !== '/mapbox-gl-js/api/' && (
```
danswick, line 27: Prettier gremlin

docs/components/quickstart-map.js
```diff
@@ -0,0 +1,103 @@
...
+<script>
+    var map = new mapboxgl.Map({
+        container: 'map', // container id
+        style: 'mapbox://styles/mapbox/streets-v11', // stylesheet location
+        center: [174.790, -36.854], // starting position [lng, lat]
+        zoom: 11 // starting zoom
+    });
+</script>`;
+
+class QuickstartMap extends React.Component {
```
danswick, line 21: I quickly adapted this from the example code components. It would be nice to make a more dedicated embedded map component if folks like having the live example here.

docs/components/quickstart-map.js
```diff
@@ -0,0 +1,103 @@
...
+
+        if (!this.iframe) return;
+        const doc = this.iframe.contentWindow.document;
+        doc.open();
+        doc.write(this.exampleHtml);
+        doc.close();
+
+        MapboxPageShell.afterUserCheck(() => {
+            this.setState({
+                token: MapboxPageShell.getUserPublicAccessToken() // eslint-disable-line
```
danswick, line 40: I think this `eslint-disable` rule might not be needed in more recent versions of eslint? https://github.com/yannickcr/eslint-plugin-react/issues/1964

docs/components/quickstart-map.js
katydecorah: Instead of repeating code from example_code.js, let's embed the "Simple map" example instead. We should be able to use the ExampleCode component and load in the html from `../example/simple-map.html`:

docs/pages/api/index.md
```diff
@@ -29,15 +31,35 @@ contentType: API
...
+}}
+
 To get started, you need an [access token](https://docs.mapbox.com/help/how-mapbox-works/access-tokens/) and a [style URL](https://docs.mapbox.com/help/glossary/style-url/). You can choose from one of our [professionally designed styles](https://docs.mapbox.com/api/maps/#styles) or create your own using [Mapbox Studio](https://www.mapbox.com/studio/).
 
 {{
 <Quickstart />
 }}
 
 
+## Reading this documentation
```
katydecorah, line 48: This is SO good

docs/pages/api/index.md
```diff
@@ -29,15 +31,35 @@ contentType: API
...
+
+This documentation is divided into several sections: 
+
+* **Map**. The `Map` object represents the map on your page. It lets you access methods and properties for interacting with the map's style and layers, respond to events, and manipulate the user's perspective with the camera.
```
katydecorah, line 52: Thoughts on adding links to each page from each bullet?

docs/pages/api/index.md
```diff
@@ -1,14 +1,16 @@
 ---
 title: Introduction
-description: Introduction to Mapbox GL JS, a JavaScript library that uses WebGL to render interactive maps from vector tiles and Mapbox styles.
+description: The Mapbox GL JS API documentation to render interactive maps from vector tiles and Mapbox styles.
+contentType: API
+lanaguage:
```
katydecorah, line 5: I think I introduced this typo a while back 🤦 
```suggestion
language:
```

docs/pages/api/index.md
```diff
@@ -29,15 +31,35 @@ contentType: API
 
 Mapbox GL JS is a JavaScript library that uses WebGL to render interactive maps from [vector tiles](https://docs.mapbox.com/help/glossary/vector-tiles/) and [Mapbox styles](/mapbox-gl-js/style-spec/). It is part of the Mapbox GL ecosystem, which includes [Mapbox Mobile](https://www.mapbox.com/mobile/), a compatible renderer written in C++ with bindings for desktop and mobile platforms.
 
+
 ## Quickstart
 
+{{
+    <QuickstartMap />
```
colleenmcginnis, line 38: I think the map should come after the instructions / code snippets.

docs/pages/api/index.md
colleenmcginnis: I had the same thought (and could also go either way)!

docs/pages/api/index.md
```diff
@@ -29,15 +31,35 @@ contentType: API
...
+* **Properties and options**. This section describes Mapbox GL JS's global properties and options that you might want to access while initializing your map or accessing information about its status.
+* **Markers and controls**. This section describes the user interface elements that you can add to your map. The items in this section exist outside of the map's `canvas` element.
+* **Geography and geometry**. This section includes general utilities and types that relate to working with and manipulating geographic information or geometries.
+* **User interaction handlers**. The items in this section relate to the ways in which the map responds to user input.
+* **Sources**. This section describes the source types Mapbox GL JS can handle in addition to the ones described in the [Mapbox Style Specification](https://docs.mapbox.com/mapbox-gl-js/style-spec/).
+* **Events**. This section describes the different types of events that Mapbox GL JS can raise.
+
+Each section describes classes or objects as well as their **properties**, **parameters**, **instance members**, and associated **events**. Many sections will also include inline code examples and related resources.
```
colleenmcginnis, line 60:
```suggestion
Each section describes classes or objects as well as their **properties**, **parameters**, **instance members**, and associated **events**. Many sections also include inline code examples and related resources.
```

docs/pages/api/index.md
danswick: That's how I originally had it configured, but changed it to this order because 1) it puts the visual "above the fold" and 2) gives the user a preview of what they will get by following the quickstart.

docs/components/quickstart-map.js
danswick: Oh hm, this will place the entire example page in the iframe. Since our example iFrames are created dynamically rather than referencing a page hosted at a set URL, I don't know of a way to make this work. Even just using the `ExampleCode` component still plops the code snippet/JSFiddle links/etc below the map. I could refactor a bit to separate out the map bit from the code snippet part of the `ExampleCode` component so we could at least reuse a component. Other ideas?

katydecorah: @danswick Ahh my mistake, I thought we wanted the code, but now I see that doesn't makes sense for this page. Maybe we could add a prop to ExampleCode to not display the CodeSnippet.

docs/components/quickstart.js
```diff
@@ -12,17 +12,30 @@ class Quickstart extends React.Component {
         };
     }
 
+    mapOptions = `{
+    container: '<your HTML element id>', //eslint-disable-line
+    style: 'mapbox://styles/mapbox/streets-v11', // stylesheet location
+    center: [-74.5, 40], // starting position [lng, lat]
+    zoom: 9 // starting zoom
+}`;
```
danswick, line 20: There were a couple of issues with the quickstart snippet:

1. It was using a really old version of mapbox streets.
2. It didn't match the embedded simple map preview.
3. The map options code is used in the CDN and bundler version of the copyable snippet.

Hopefully a single map options variable will help keep this up to date. Ideally, we would replace the mapbox streets style version with a constant, but I don't think we do that anywhere else in this repo. I may just be missing something!

docs/components/quickstart.js
```diff
@@ -12,17 +12,30 @@ class Quickstart extends React.Component {
...
+
     componentDidMount() {
         MapboxPageShell.afterUserCheck(() => {
             this.setState({
                 userAccessToken:
                     MapboxPageShell.getUserPublicAccessToken() ||
-                    `<your access token here>`
+                    `\n\t// TO MAKE THE MAP APPEAR YOU MUST\n\t// ADD YOUR ACCESS TOKEN FROM\n\t// https://account.mapbox.com`
```
danswick, line 27: Match the louder and more descriptive token replacement warning recently added to examples.

docs/components/quickstart.js
```diff
@@ -12,17 +12,30 @@ class Quickstart extends React.Component {
...
         });
     }
 
     renderCdn() {
+        const cdnMapHtml = `
+<div id='map' style='width: 400px; height: 300px;'></div>
+<script>
+mapboxgl.accessToken = '${this.state.userAccessToken}';
+var map = new mapboxgl.Map(${this.mapOptions});
+</script>`;
```
danswick, line 38: We can't use the `simple-map.html` code here directly because: 

1. It doesn't come with any CSS. This is possibly ok, but i think the spirit of this quickstart is to make the code as easy to drop into place as possible.
2. The `accessToken` is added programmatically. 

We could transform the HTML from the example, but I am not convinced the benefit would be worth the effort and complexity.

PR #316, opened 2020-05-12
[api-pages] rename md function -> toHtml; move toHtml and formatType to formatters

katydecorah:
**This branch merges into `api-ref-main`**.

This PR adds a little clean up:
- renamed `md` function to `toHtml` (since it outputs html and not markdown)
- moved `toHtml` and `formatType` functions to formatters.js (since they are dependencies)
- imported functions in sub components rather than pass them as props


docs/components/api/item-contents.js
```diff
@@ -9,36 +9,10 @@ import Returns from './returns';
 import Throws from './throws';
 import Examples from './examples';
 import Related from './related';
-import formatters from '../../util/formatters';
+import { toHtml } from '../../util/formatters';
 import empty from '../../util/empty';
 
 class ApiItemContents extends React.Component {
-    md = (ast, inline) => {
```
danswick, line 16: phew! much nicer. :dusty_stick:

docs/util/formatters.js
```diff
@@ -1,3 +1,5 @@
+/* eslint react/no-danger: 0 */
```
danswick, line 1: Interesting. Why wasn't eslint unhappy about this when it the `toHtml` (né `md`) functionality was in `item-contents`? 

edit: ah, I see that the rule is disabled in components here: https://github.com/mapbox/mapbox-gl-js-docs/blob/publisher-production/docs/components/.eslintrc#L4. I'm not sure how much of the eslint configuration is just a legacy of the old docs setup, but do you think we would want to move this rule up the `/docs/.eslintrc`? Or is it better to keep it specific to components?

katydecorah: Wow thanks for catching that, I'll do some eslint cleaning in this PR!

PR #318, opened 2020-05-15
[api-pages] Add page descriptions and page descriptions component

danswick:
Adds intro text to each page: 

- ~~Adds a new small component, `ApiPageDescription`. This might not be necessary since it's really pretty tiny and I don't think it actually needs its own special formatting.~~ Ok, yeah, that component was worthless and has been excised. 
- Scoots some of the ApiPageItems `const`s from inside the render function to higher up in the place, allowing the `renderDescription` function to work for the `Section` component outside the scope of the render function.
- Add some (pretty boring) intro text to `documentation.yml`, including some basic markdown formatting.


docs/components/api/api-page-items.js
```diff
@@ -5,40 +5,57 @@ import ApiItemContents from './item-contents';
...
-                    <GroupedSection
-                        key={child.name}
-                        location={this.props.location}
-                        members={child.members}
-                        name={child.name}
-                    />
+                    <React.Fragment key={child.name}>
+                        {this.renderDescription()}
+                        <GroupedSection
+                            key={child.name}
```
katydecorah, line 45: I think we can remove `key` here since the key iterator is now on the wrapping React.Fragment

docs/components/api/api-page-items.js
```diff
@@ -5,40 +5,57 @@ import ApiItemContents from './item-contents';
...
+                            name={child.name}
+                        />
+                    </React.Fragment>
                 );
             } else {
-                return <Section key={child.name} {...this.props} {...child} />;
+                return (
+                    <React.Fragment key={child.name}>
+                        {this.renderDescription()}
+                        <Section key={child.name} {...this.props} {...child} />
```
katydecorah, line 56: I think we can also remove the key on Section here, too.

docs/components/api/api-page-items.js
```diff
@@ -5,40 +5,57 @@ import ApiItemContents from './item-contents';
...
 export default class ApiPageItems extends React.Component {
-    render() {
-        const pageDocSource = apiFilterItems(this.props.pageTitle);
-        const children = pageDocSource[0].members.static;
+    pageDocSource = apiFilterItems(this.props.pageTitle);
+    children = this.pageDocSource[0].members.static;
+    renderDescription = () => {
+        const description = this.pageDocSource[0].description || false;
+        if (description)
+            return <React.Fragment>{toHtml(description)}</React.Fragment>;
```
katydecorah, line 18: I think you could remove the React.Fragment wrapper and return:

```suggestion
            return toHtml(description);
```

docs/components/api/api-page-items.js
```diff
@@ -5,40 +5,57 @@ import ApiItemContents from './item-contents';
...
             return (
-                <SingleSection
-                    key={children[0].name}
-                    {...children[0]}
-                    {...this.props}
-                />
+                <React.Fragment>
+                    {this.renderDescription()}
+                    <SingleSection
+                        key={this.children[0].name}
```
katydecorah, line 32: I know this was already in the diff, but we could remove the `key` too.

docs/documentation.yml
```diff
@@ -6,6 +6,8 @@ toc:
 
   - name: Properties and options
     page: 'properties'
+    description: |
+      Mapbox GL JS's global properties and options that you might want to access while initializing your map or accessing information about its status.
```
katydecorah, line 10: Small suggestion, take or leave!

```suggestion
      Mapbox GL JS's global properties and options that you can access while initializing your map or accessing information about its status.
```

docs/components/api/api-page-items.js
```diff
@@ -5,40 +5,57 @@ import ApiItemContents from './item-contents';
...
+                        {...this.children[0]}
+                        {...this.props}
+                    />
+                </React.Fragment>
             );
         }
-        return children.map(child => {
+        return this.children.map(child => {
```
katydecorah, line 39: @danswick Whoops! I just realized that the description is getting iterated before _every_ section, when I think it would only show once at the top of the page. 

> image

I think for the GroupSection and Section varieties, we'll want to add the description before we loop through children.

danswick: Wow that's gonna be a problem 😂

docs/components/api/api-page-items.js
danswick: React complains about this one :(

danswick: OH i see that you're talking about the one on 45 and not the one on 42 🤦

docs/components/api/api-page-items.js
danswick: OK! The fix for this also addressed the `key` issues as a bonus 😅

PR #319, opened 2020-05-15
[api-pages] Remove grouped section layout

katydecorah:

This PR remove the grouped section layout we applied to the new Properties page. We found that the sections didn't really help in organizing the content and instead made each item harder to find in the sidebar. 

I also noticed that "Returns" was in the wrong order in the sidebar and fixed that here too.

@danswick I used the order that's currently on production. Look ok? Should we alphabetize? I don't really have a preference.


docs/components/api/api-page-items.js
```diff
@@ -84,55 +65,22 @@ Section.propTypes = {
...
-};
-
 class SingleSection extends React.Component {
     render() {
+        const child = this.props;
         return (
             <React.Fragment>
                 <div className="mt-neg30">
-                    <Github section={this.props} />
+                    <Github section={child} />
```
danswick, line 74: I can't remember if we already talked about this, but all of the github paths have `:mapbox` (with colon) in the URL, causing them to 404. I think this is unrelated to this PR, but I just noticed it 👀

katydecorah: @danswick I'm having trouble reproducing, can you point me to the exact link that you saw 404? I'm the meantime, I'm going to merge and we can follow up on this in another PR!

PR #323, opened 2020-05-18
Use @mapbox/mapbox-gl-style-spec package

### LINKED ISSUE #298: Support pinning to latest style-spec tags as well as latest gl-js tags
Sometimes, style-spec updates happen outside of the normal `mapbox-gl-js` release schedule. We should be able to update the style spec without getting `mapbox-gl-js` out of sync.

cc @asheemmamoowala

katydecorah:
This PR switches from using the mapbox-gl submodule style spec version to the package `@mapbox/mapbox-gl-style-spec`. 

With this change:
- We can pin the style-spec version without needing to bump the mapbox-gl submodule. Fixes #298.
- We can remove an extra build step to install style-spec dependencies. Fixes #245 

## Holding

- [x] Waiting for submodule bump happening in #321. This will make the tests pass.
- [x] `distance` was added in https://github.com/mapbox/mapbox-gl-js/pull/9655 which was included in mapbox-gl@1.10.1 (or 1.10.0) but the style-spec hasn't had a release yet that includes `distance`. We should wait until that release happens.

cc @danswick @asheemmamoowala


asheemmamoowala:
>  distance was added in mapbox/mapbox-gl-js#9655 which was included in mapbox-gl@1.10.1 (or 1.10.0) but the style-spec hasn't had a release yet that includes distance. We should wait until that release happens.

@karimnaaji Could you also make a patch release of the style-spec package since there were a number of changes there.

karimnaaji:
Yes! I'll publish this so the docs can be pulled in. https://github.com/mapbox/mapbox-gl-js/pull/9706

karimnaaji:
Published in https://github.com/mapbox/mapbox-gl-js/releases/tag/style-spec%4013.15.0.

PR #326, opened 2020-05-22
[api-pages] exploring adding an API Reference search to the sidebar

katydecorah:
@danswick Based on feedback we've received so far in #250, it seems like an ability to finely search the API reference would be helpful to supplement the loss of being able to ctrl+f. 

I started exploring the idea of adding an API Reference search. The search will look for matches in the `name`, `namespace`, and `description` and directly links to the item selected.

2020-05-22 at 1 18 PM

A few changes to make this happen:
- This PR adds a new batfish dataSelector, `apiSearch`, to process api.json and build a cleaned up JSON for the search.
- A few files needed to be converted to commonjs so they could be used in the new dataSelector.

If we like this plan, I think it would be great to eventually add this type of search to the Style Spec sidebar, too.

It's not ideal to add another search option to the page when we already have a docs-wide search. But, we do already use an input filter on the Example and Plugins page. Since how we're using Swiftype right now, it cannot deep link to sections (ref https://github.com/mapbox/documentation/issues/204) - I think this additional search option could be holdover until we're able to evaluate making adjustments to how Swiftype ingests our content.


docs/components/api/search.js
```diff
@@ -0,0 +1,137 @@
...
+                                <label
+                                    className="hide-visually"
+                                    {...getLabelProps()}
+                                >
+                                    Search API Reference
+                                </label>
+                                <input
+                                    placeholder="Search API Reference"
+                                    className="input bg-white"
+                                    {...getInputProps({
```
danswick, line 65: oh interesting! I will need to read do some homework on Downshift 👀

PR #334, opened 2020-05-25
Add mapbox-gl-indoorequal plugin

francois2metz:
- plugin: https://github.com/indoorequal/mapbox-gl-indoorequal
- website: https://indoorequal.org/


francois2metz:
Is there is anything that I can do to make it mergeable? I was wondering if using a mapbox basemap is mandatory. Feel free to reach me out privately if needed.

PR #335, opened 2020-05-28
Update docs-page-shell

katydecorah:
- Adds MTS to Maps navigation


katydecorah:
Don't need this PR afterall, the deploy script can handle it.

PR #341, opened 2020-06-04
Update examples that use icons from default sprites

HeyStenson:
This PR updates examples that use icons from default sprites so that they either use custom icons added using `addImage` (or some other solution, depending on the example). @mapbox/docs team is doing this because we've received a number feedback tickets from Mapbox users who are confused about where these icons are coming from and how to find the complete list. Since finding that list is not a straightforward procedure, nor is it the most stable way to build, we are redoing examples that use that approach:

- [x] https://docs.mapbox.com/mapbox-gl-js/example/geojson-markers/
- [x] https://docs.mapbox.com/mapbox-gl-js/example/center-on-symbol/
- [x] https://docs.mapbox.com/mapbox-gl-js/example/popup-on-hover/

Screen Shot 2020-06-03 at 5 45 55 PM

Screen Shot 2020-06-04 at 3 49 59 PM

Screen Shot 2020-06-08 at 1 37 42 PM

**Edit:** I'm going to address these examples in a different PR since it's taking me a while to get to them and I'd like to get the ones that I've already changed updated:

- https://docs.mapbox.com/mapbox-gl-js/example/fallback-image/
- https://docs.mapbox.com/mapbox-gl-js/example/filter-markers-by-input/
- https://docs.mapbox.com/mapbox-gl-js/example/filter-markers/
- https://docs.mapbox.com/mapbox-gl-js/example/popup-on-click/
- https://docs.mapbox.com/mapbox-gl-js/example/change-case-of-labels/


andrewharvey:
Added bonus is that these markers are much easier actually see, click and hover compared to before. :rocket:

docs/pages/example/popup-on-click.html
```diff
@@ -14,131 +14,131 @@
     });
 
     map.on('load', function() {
```
asheemmamoowala, line 16: One of the concepts shown in this example is data-driven styling of icon images., which I was not able to find in any other examples. If we remove the data-driven styling here, a new example to show how this is done would be helpful to explain how icons from a sprite can be referenced using DDS and/or expressions.

HeyStenson: Thanks @asheemmamoowala! I'm reverting the changes to this file in this PR and will update this example in a way that still uses DDS in a different PR.

HeyStenson:
Adding the custom marker in a [separate PR](https://github.com/mapbox/mapbox-gl-js-docs/pull/359).

docs/pages/example/popup-on-click.js
```diff
@@ -1,6 +1,6 @@
 /*---
 title: Display a popup on click
-description: 'When a user clicks a symbol, show a [`Popup`](/mapbox-gl-js/api/#popup) containing more information. The symbols are from the [Maki](https://labs.mapbox.com/maki-icons/) symbol set used in the Mapbox Streets style.'
+description: 'When a user clicks a custom marker, show a [`Popup`](/mapbox-gl-js/api/#popup) containing more information.'
```
katydecorah, line 3: This example isn't using the same custom marker as the others, did you mean to update this description?

HeyStenson: Hoo boy, sure didn't (or at least didn't mean to update it in this PR 😬 ). Thanks, I'll update!

PR #356, opened 2020-06-24
include css import in quickstart module bundler

andrewharvey:
[Webpack has a css-loader module](https://webpack.js.org/loaders/css-loader/) which makes importing GL JS as easy as:

```js
import mapboxgl from 'mapbox-gl'
import 'mapbox-gl/dist/mapbox-gl.css'
```

The [quickstart](https://docs.mapbox.com/mapbox-gl-js/api/#quickstart) module bundler example imports the css via HTML `<link>` which is good to show as not everyone will have something like `css-loader` setup, but I think it's still worth showing the import css as an option as many starter kits and boilerplates already have this setup.

Here's what this PR looks like

Screenshot from 2020-06-24 11-36-01

Personally I prefer to import the css this way as it ensures the JS and CSS are kept in sync when updating dependencies through npm/yarn. There's been a few issues reported by uses in GL JS where the root cause was out of sync CSS and JS and I wonder if import js / link css and then upgrading through yarn or npm but not the HTML link were cause.


katydecorah (approved):
Thanks for your patience on this, it's a great add. I made a small copyedit and moved the two CSS code snippets together:

image

PR #372, opened 2020-07-23
Add example that demonstrates how to display a line that crosses 180th meridian 

malwoodsantoro:
image

This example seeks to further explain the concept described within this comment https://github.com/mapbox/mapbox-gl-js/issues/3250#issuecomment-294887678.


katydecorah:
@malwoodsantoro We made some changes to the examples file format in #374 - instead of JavaScript files we are now using markdown files. I updated your branch to match the new system. Let me know if you have any questions!

katydecorah:
@malwoodsantoro hi again :) I updated this branch with a newly required frontmatter item (ref #377) - No actions on your end!

malwoodsantoro:
Thank you for both updates, @katydecorah! ✨

docs/pages/example/line-across-180th-meridian.html
```diff
@@ -0,0 +1,84 @@
...
+                    'type': 'Feature',
+                    'properties': {},
+                    'geometry': {
+                        'type': 'LineString',
+                        // To draw a line across the 180th meridian,
+                        // if the longitude of the second point minus
+                        // the longitude of original (or previous) point is > 180,
+                        // subtract 360 from the longitude of the second point.
+                        // If it is less than 180, add 360 to the second point.
+                        'coordinates': [
```
ryanhamley, line 60: This doesn't account for if the difference is exactly 180. I think from playing with it, the algorithm should be `second - original >= 180`.

I'm also wondering if a function to do this would be more explicit and show how to update a layer on the fly. If you add a layer that's in a tile or style, you may have to update it at runtime.

Something like this works:
```
function createGeometry(doesCrossAntimeridian) {
  const geometry = {
    'type': 'LineString',
    'coordinates': [
      [360, 35.67514],
      [180, -16.59408]
    ]
  };

  if (doesCrossAntimeridian) {
    const startLng = geometry.coordinates[0][0];
    let endLng = geometry.coordinates[1][0];

    if (endLng - startLng >= 180) {
      geometry.coordinates[1][0] -= 360;
    } else if (endLng - startLng < 180) {
      geometry.coordinates[1][0] += 360;
    }
  }

  return geometry;
}
```

 You'd use it like

```
'data': {
    'type': 'Feature',
    'properties': {},
    'geometry': createGeometry()
}
```

or `'geometry': createGeometry(true)` to set the line across the 180th meridian. It's just a suggestion. I'd keep the comment you have explaining the algorithm, but I think the code might make it easier to understand.

PR #374, opened 2020-07-28
Switch examples from `js` file format to `md`

katydecorah:
This PR will switch all example from using `js` file format to `md`. This change will allow these 100+ example files to take advantage of the docs suite of remark and copy-related tests.

This branch is live on staging.

Fixes https://github.com/mapbox/documentation/issues/516


.remarkrc.js
```diff
@@ -3,7 +3,8 @@ const config = {
     siteBasePath: 'mapbox-gl-js',
     pages: 'docs/pages/',
     constants: 'docs/constants.json',
-    ignoreLinks: 'conf/ignore-links.json'
+    ignoreLinks: 'conf/ignore-links.json',
+    descriptionLength: 1200
```
katydecorah, line 7: This overrides the default. Right now the site is using the `description` field to populate the content of example pages and there are a few pages with very big descriptions. Eventually we will shift this content into the content of the markdown file.

docs/data/plugins.json
```diff
@@ -126,7 +126,7 @@
     "mapbox-gl-supported": {
       "website": "https://github.com/mapbox/mapbox-gl-supported",
       "description": "Determines if the current browser supports Mapbox GL JS.",
-      "example": "mapbox-gl-supported"
+      "example": "check-for-support"
```
katydecorah, line 129: The mapbox-gl-support example redirects to check-for support.

docs/components/example_code.js
```diff
@@ -14,7 +14,7 @@ const viewport = `<meta name="viewport" content="initial-scale=1,maximum-scale=1
 const css = `\tbody { margin: 0; padding: 0; }
 \t#map { position: absolute; top: 0; bottom: 0; width: 100%; }`;
 
-export default class ExampleCode extends React.Component {
+export default class Example extends React.Component {
```
colleenmcginnis, line 17: Should we also update the filename from `docs/components/example_code.js` to `docs/components/example.js`, too?

docs/pages/examples.js
```diff
@@ -29,10 +29,10 @@ class ExamplesLandingPage extends React.PureComponent {
             const cardsForTopic = examples
                 .filter(example => example.tags.indexOf(topic) > -1)
                 .map((example, index) => {
-                    const filename = example.pathname.split('/')[3];
+                    const thumbnail = example.thumbnail;
```
colleenmcginnis, line 32: Do we really need the thumbnail field or could we derive this from `location.pathname` since the image file will always have the same name? This seems like it could introduce more risk of human error.

katydecorah: In dr-ui layouts, we'll need the `thumbnail` for generating the images for cards (this is a pattern shared by other repos). 

> This seems like it could introduce more risk of human error.

There are also tests in place to assert that the thumbnail is a valid image that exists in this repo.

docs/components/example_code.js
katydecorah: 👍 renamed!

PR #388, opened 2020-08-13
Link to Geocoding API docs in example description

domlet:
Closes https://github.com/mapbox/mapbox-gl-js-docs/issues/387

Adding this hyperlink may help this user find their own answer in Geocoding API docs.


andrewharvey:
This is great, we should extend this to all the geocoder examples which references the Geocoding API.

domlet:
Thank you @andrewharvey – @adrianababakanian has suggested a programmatic solution for references in GL JS examples and we hope to move forward with this in 2020: https://github.com/mapbox/documentation/issues/321

PR #397, opened 2020-08-24
Modify example so it clearly uses a custom style

domlet:
Closes: https://github.com/mapbox/documentation/issues/488


docs/pages/example/custom-style-id.html
```diff
@@ -2,7 +2,7 @@
 <script>
     var map = new mapboxgl.Map({
         container: 'map', // container id
-        style: 'mapbox://styles/mapbox/dark-v10', //hosted style id
+        style: 'mapbox://styles/examples/cke97f49z5rlg19l310b7uu7j', // custom style url id
```
katydecorah, line 5: Maybe add where to find the style url?

```suggestion
        style: 'mapbox://styles/examples/cke97f49z5rlg19l310b7uu7j', // custom style url from https://studio.mapbox.com/
```

katydecorah: This map. OMG. amazing.

Issue #398, opened 2020-08-26
Directions plugin doesn't work in IE11


katydecorah:
The directions plugin is not compatible with IE11:

> ![image](https://user-images.githubusercontent.com/2180540/91306066-0d192700-e77a-11ea-8fb1-54214f3c6d76.png)

We should add a polyfill to any example using the Directions plugin to make it compatible with IE 11. We currently do this with the geocoding plugin:

https://github.com/mapbox/mapbox-gl-js-docs/blob/d4d7909a286beae1de454ed8f8a068760195776d/docs/pages/example/mapbox-gl-geocoder-no-map.html#L7-L9

cc @malwoodsantoro

katydecorah:
mapbox-gl@2.0.0 no longer supports IE 11.

PR #399, opened 2020-08-26
Update/add IE11 polyfills for directions and geocoder plugins

katydecorah:
This PR configures polyfills so that the directions and geocoder plugins used in examples will be compatible with IE 11:

- adds babel/polyfill to the directions plugin example so that it will be compatible with IE 11.
- updates the geocoder plugins to also use babel/polyfill. Currently each geocoder example uses the promise polyfill, but several examples required more covered (for example startsWith).

Fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/398

cc @malwoodsantoro @scottsfarley93

### DIFF
--- a/docs/pages/example/mapbox-gl-geocoder-no-map.html
+++ b/docs/pages/example/mapbox-gl-geocoder-no-map.html
@@ -1,12 +1,13 @@
+<!-- Polyfill scripts required to use Mapbox GL Geocoder in IE 11 -->
+<script src="https://cdn.jsdelivr.net/npm/@babel/polyfill@7.10.4/dist/polyfill.min.js"></script>
+<script src="https://cdn.jsdelivr.net/npm/nodelist-foreach-polyfill@1.2.0/index.min.js"></script>
+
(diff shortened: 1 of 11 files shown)


docs/pages/example/mapbox-gl-geocoder-no-map.html
```diff
@@ -1,12 +1,13 @@
+<!-- Polyfill scripts required to use Mapbox GL Geocoder in IE 11 -->
+<script src="https://cdn.jsdelivr.net/npm/@babel/polyfill@7.10.4/dist/polyfill.min.js"></script>
+<script src="https://cdn.jsdelivr.net/npm/nodelist-foreach-polyfill@1.2.0/index.min.js"></script>
```
katydecorah, line 3: In my testing, this was the only example that ran into issue with NodeList forEach.

malwoodsantoro:
Thank you, @katydecorah. I have two PRs open to update both of the plugin repo READMEs once this is merged https://github.com/mapbox/mapbox-gl-directions/pull/262/files and https://github.com/mapbox/mapbox-gl-geocoder/pull/394 👍

katydecorah:
Now that mapbox-gl@2.0.0 no longer supports IE 11, I think this PR is no longer needed. Closing.

PR #411, opened 2020-08-29
added some plugins for mapbox-gl-js

JinIgarashi:
I developed some plugins for [my project](https://narok.water-gis.com/). Is it possible to add them into your plugins page?
- mapbox-gl-legend: https://github.com/watergis/mapbox-gl-legend
- mapbox-gl-popup: https://github.com/watergis/mapbox-gl-popup
- mapbox-gl-print: https://github.com/watergis/mapbox-gl-print


docs/data/plugins.json
```diff
@@ -36,10 +36,22 @@
       "website": "https://github.com/el/infobox-control",
       "description": "Adds a control to display an infobox or a gradient."
     },
+    "mapbox-gl-legend": {
+      "website": "https://github.com/watergis/mapbox-gl-legend",
+      "description": "adds legend control which is able to create legend panel from mapbox style to mapbox-gl."
```
ryanhamley, line 41: Let's reword this: `Adds a control that shows a legend generated from the map style`

docs/data/plugins.json
```diff
@@ -36,10 +36,22 @@
...
+    },
     "mapbox-gl-opacity": {
       "website": "https://github.com/dayjournal/mapbox-gl-opacity",
       "description": "Make multiple tile layers transparent."
     },
+    "mapbox-gl-popup": {
```
ryanhamley, line 47: This one isn't really a control. I'm inclined to not include this one since we already have popups built into the library and the name and functionality of this is somewhat confusing and it allows for a fairly limited use case.

docs/data/plugins.json
ryanhamley: This is cool. My humble suggestion is to make the legend show by default and make the minimized button view a configurable state, rather than the default. That's not something you need to do before we include the plugin though.

docs/data/plugins.json
```diff
@@ -36,10 +36,22 @@
...
+      "website": "https://github.com/watergis/mapbox-gl-popup",
+      "description": "Adds popup control which can show attributes of targeted layers to mapbox-gl."
+    },
+    "mapbox-gl-print": {
```
ryanhamley, line 51: This is an interesting plugin. I'm worried though that the name doesn't match the functionality. This doesn't seem to actually initiate a print. What it does is useful, but a better name might be something like `mapbox-gl-export`. I'd suggest either adding the print functionality or changing the name.

ryanhamley (changes requested):
Hi @JinIgarashi thanks for your work and interest!

My take on these three plugins:

1. mapbox-gl-legend: very cool! definitely want to add this to the list

2. mapbox-gl-popup: i think this needs to be developed more. the use case that it allows for is limited and the name is somewhat confusing since we already have popups in the library

3. mapbox-gl-print: this works well but the name is confusing since it doesn't actually print. i'd suggest adding that functionality or changing the name

docs/data/plugins.json
JinIgarashi: @ryanhamley That's true. Let me remove this plugin from pull request.

docs/data/plugins.json
JinIgarashi: @ryanhamley Sure, I agree that.export is much better than print. I will try to rename to be `mapbox-gl-export`, then I will apply to this pull request.

docs/data/plugins.json
JinIgarashi: @ryanhamley Thank you for your suggestion! I will try to add that function. Could you add issue in [my repository](https://github.com/watergis/mapbox-gl-legend)?

JinIgarashi:
@ryanhamley 
Thank you for your comment! I will work on my pull request as you suggested. Then let you know when it will be ready to merge.

> You'll also need to update your branch with the newest commits from `publisher-production`
Sure, I will update my branch to be the latest.

JinIgarashi:
Hi @ryanhamley
I have updated my branch and modified as you requested. Could you merge my plugins - `mapbox-gl-legend` and `mapbox-gl-export`? Please let me know if you have any additional suggestion.

docs/data/plugins.json
```diff
@@ -27,6 +27,10 @@
       "description": "Adds support for drawing and editing features on Mapbox GL JS maps.",
       "example": "mapbox-gl-draw"
     },
+    "mapbox-gl-export": {
+      "website": "https://github.com/watergis/mapbox-gl-export",
+      "description": "Adds a control that exports PDF and image."
```
ryanhamley, line 32: It just exports `PNG` correct? Can we update this to be specific? `Adds a control that exports the map as a PDF or PNG`

ryanhamley (changes requested):
Thanks for the quick turnaround on the changes! Looks great. I just had one suggestion on the description for the export plugin.

docs/data/plugins.json
JinIgarashi: I used a word of "image" because I might add another format of image, but it only export PNG image currently. so no problem to specify. I just updated description as you suggested.

JinIgarashi:
@ryanhamley I updated description of export plugin as you suggested. Could you check it?

ryanhamley (approved):
Thanks!

PR #417, opened 2020-09-19
Change map center and update thumbnail image to match new style

domlet:
For this example: https://docs.mapbox.com/mapbox-gl-js/example/custom-style-id/

old thumb:

Screen Shot 2020-09-18 at 6 23 45 PM

new thumb:

Screen Shot 2020-09-18 at 6 23 22 PM


HeyStenson (approved):
Looks great! 🍊

PR #423, opened 2020-09-29
updated mapbox minimap plugin

abdulk1:
Updated mapbox minimap plugin to a different one. The new one is actively maintained and comes with types.


danswick:
Thanks, @abdulk1. It looks like the current minimap is pretty out of date and seems only to support up to Mapbox GL JS 0.54. I think we would be ok updating this, but just want to tag in @mapbox/gl-js to confirm.

PR #428, opened 2020-10-07
Update data join example to use mapbox countries tileset

planemad:
Ref https://github.com/mapbox/documentation/issues/555

- Updated to use `mapbox-country-boundaries-v1` tileset
- Used the Human Development Index data filtered for only Europe for the data join

Screen Shot 2020-10-07 at 8 14 09 PM

<details>

Stale version

- Updates the data-join example to use the `mapbox.country-boundaries-v1` default tileset
- Use Press Freedom Index data https://rsf.org/en/ranking_table
- Join data using feature state rather than building a color mapping for each individual feature

**Questions**
- Is the dataset used ok, should we use something else or a smaller subset instead of data for all countries?
- Since the mechanism of the data join is different, should we instead make this a seperate example?

</details>

cc @HeyStenson @JB2013 @ajashton


planemad:
Screen Shot 2020-10-07 at 6 26 47 PM

planemad:
- Reworked to retain the join with a match expression
- Used the Human Development Index data which might be less controversial than the Press Freedom Index, and limited to just Europe.

Screen Shot 2020-10-07 at 8 14 09 PM

HeyStenson (approved):
LGTM! 🍊

planemad:
🥳

PR #429, opened 2020-10-08
Add example to show how to join data using feature state

planemad:
Ref: https://github.com/mapbox/documentation/issues/555#issuecomment-705795649

This example is based on the existing data-join example https://github.com/mapbox/mapbox-gl-js-docs/pull/428 but uses feature-state to join the json data directly to the vector tiles.

This is the data join mechanism recommended with boundaries tilesets https://docs.mapbox.com/help/tutorials/data-joins-with-mapbox-boundaries/#feature-state

### Open questions

- How do we better distinguish this from the data-join example so that it is not confusing for users?

Screen Shot 2020-10-08 at 4 33 45 PM

cc @HeyStenson @JB2013 @ajashton


planemad:
From slack

Screen Shot 2020-10-08 at 10 46 35 PM

One use case that comes up from boundaries users is to join custom labels to the shapes, looks like this is not possible using the feature state join in GL.

cc @asheemmamoowala @jbranigan

asheemmamoowala:
> How do we better distinguish this from the data-join example so that it is not confusing for users?

We can do a couple things to separate these two examples. First focus this one on using `paint` properties, and the other on `layout` properties. And add a description here to clarify that `feature-state` only works with `paint` properties, with a link to the other example for joining with `layout` properties.


> One use case that comes up from boundaries users is to join custom labels to the shapes, looks like this is not possible using the feature state join in GL.

That's right, the `feature-state` data is not available during the layout phase, so it cannot be used to change the label on a shape.

PR #444, opened 2020-11-11
Fix typo in custom style id example

domlet:
I left some goofiness ("This examples uses") in the past PR, so here's a fix.

Screen Shot 2020-11-10 at 5 10 26 PM


HeyStenson (approved):
YIKE. Sorry I missed this the first go-round! 🙈 Edits look good!

PR #446, opened 2020-11-18
[Info arch] Update to dr-ui 2.0.0

danswick:
This PR updates dr-ui@2.0.0.

This update introduces layout, navigation, and design changes for the site to help improve information architecture. See [changelog](https://github.com/mapbox/dr-ui/blob/info-arch-main/CHANGELOG.md#info-arch) for a list of all changes.

## Checklist

* [x] Update dr-ui, remark-lint-mapbox, sentry `npm install @mapbox/dr-ui@2.0.0-beta.0 @mapbox/remark-lint-mapbox@latest @sentry/browser@latest`
* [x] Replace “layout: accordion” with “layout: page”
* [x] Remove “hideTitle: true” on all pages.
* [x] Move OverviewHeader props to frontMatter. [Example](https://github.com/mapbox/dr-ui/pull/349/files#diff-9042dfe766cf312244f367aa1add6fe2714f6820226936541016b1e30b153267).
* Update page-shell.js:
    * [x] Replace “topics” with “filters” (Batfish helper import and PageShell prop)
* Update batfish.config.js
   * [x] Replace “buildTopics” with “buildFilters” (Batfish helper import and dataSelector)
   * [x] Remove “formatTopics” (if exists and all other dependent code including local Help JSON file.)
   * [x] Remove “create-sections” from rehypePlugins
   * [x] Update “buildNavigation” function to use named params and pass “addPages” data (if relevant).
* [x] If applicable, remove the “Help” page and add a redirect to https://github.com/mapbox/subdomain-docs/pull/110
* [x] If applicable, check the Examples page. Use `showFilters` to only display desired filters (usually exclude products).
* [x] Audit check: `npm audit` and/or `npm audit fix`
* [x] Confirm Feedback module is working
* [x] Force push to staging
* [x] Site loads in IE 11, Edge, Safari, Firefox, Android Chrome, iOS Safari

Ref https://github.com/mapbox/documentation/issues/567

PR #448, opened 2020-11-20
[info arch] Fix Property styling and attributes; move custom headings to frontMatter

katydecorah:
This PR merges into https://github.com/mapbox/mapbox-gl-js-docs/pull/446/

@danswick while reviewing #446 I noticed that the Style Spec h3 headings weren't showing up in OnThisPage (for example /mapbox-gl-js/style-spec/layers/#background does not show h3s when you scroll to it) -- this is because I introduced some typos during testing and found that for Style Spec headings we want to check to see if there are generated headings first then set markdown headings.


danswick (approved):
looks good to me!

PR #452, opened 2020-12-08
Add terrain type case statement

katydecorah:
This PR adds a style spec type case statement for `terrain`:

image


ryanhamley (approved):
thanks Katy!

PR #457, opened 2020-12-10
Use absolute path for routes.js in " Animate the camera along a path" example

### LINKED ISSUE #456: Free Camera Path Codepen error
# Issue
The line and camera animation are broken when [this example](https://docs.mapbox.com/mapbox-gl-js/example/free-camera-path/) is opened on Codepen, because it does not include a reference to the route line - the data source is referenced from a relative path. We mention this in the JavaScript comments.

image

# Suggested fix
Reference the absolute path in the Codepen template as an External Script

image

katydecorah:
This PR updates the path to routes.js in the "Animate the camera along a path" example to use an absolute path. This will make sure the code example works when copied or edited in CodePen/JSFiddle.

Fixes #456


katydecorah:
image

@segheysens Thanks for the heads up!

colleenmcginnis (approved):
Thanks @katydecorah @segheysens!

PR #466, opened 2020-12-17
Change fallback icon in Fallback icon example

HeyStenson:
fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/464

A reader pointed out that the [fallback icon example](https://docs.mapbox.com/mapbox-gl-js/example/fallback-image/) doesn't actually use the icon specified as the fallback. This seems to be because `marker-15` is not available in the Mapbox style the example uses. This PR updates it to use `rocket-15`, which is available in the style, instead. 
Screen Shot 2020-12-17 at 3 20 34 PM

@mapbox/docs to review, please 🙏


katydecorah (approved):
👩‍🎤

PR #467, opened 2020-12-17
Update Turf version and method in Animate a point example

HeyStenson:
fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/463

> This example uses Turf  version from 5 years ago The newest version has no lineDistance method.

This PR updates Turf and switches `lineDistance` for `length` in https://docs.mapbox.com/mapbox-gl-js/example/animate-point-along-route/. 

Screen Shot 2020-12-17 at 3 48 06 PM

@mapbox/docs to review, please! 🙏


katydecorah:
@HeyStenson I noticed that when the airplane reaches the destination the following error is logged to the console:

```
turf.min.js:1 Uncaught Error: coordinates is required
    at Object.r [as point] (turf.min.js:1)
    at animate ((index):117)
```

I think this is happening in the animate function, when the airplane reaches the destination the second argument to `turf.bearing` is undefined. I think if we update the animate function to make sure that the start and end coordinates exist before moving forward with the function then it should fix the error:

```js
function animate() {
  var start =
      route.features[0].geometry.coordinates[
          counter >= steps ? counter - 1 : counter
      ];
  var end =
      route.features[0].geometry.coordinates[
          counter >= steps ? counter : counter + 1
      ];
  if (!start || !end) return;

  // Update point geometry to a new position based on counter denoting
  // the index to access the arc.
  point.features[0].geometry.coordinates =
      route.features[0].geometry.coordinates[counter];

  // Calculate the bearing to ensure the icon is rotated to match the route arc
  // The bearing is calculate between the current point and the next point, except
  // at the end of the arc use the previous point and the current point
  point.features[0].properties.bearing = turf.bearing(
      turf.point(start),
      turf.point(end)
  );

  // Update the source with this new data.
  map.getSource('point').setData(point);

  // Request the next frame of animation so long the end has not been reached.
  if (counter < steps) {
      requestAnimationFrame(animate);
  }

  counter = counter + 1;
}
```

HeyStenson:
Thanks for catching that @katydecorah! 🙈  Updated.

katydecorah (approved):
⭐

PR #473, opened 2021-01-08
Switch to npm

### LINKED ISSUE #472: Switch from Yarn to npm
When we first split the docs from mapbox-gl-js repo, we attempted to also switch from yarn to npm (which is standard for docs sites). Due to an unidentified bug that caused the build to fail, we weren't able to make the switch so we stuck with Yarn. 

We identified and fixed the bug (https://github.com/mapbox/mapbox-gl-js-docs/pull/60#issuecomment-739027243) and can now safely move over to npm.

@mapbox/gl-js @mapbox/docs Any objections to switching from yarn to npm for this repository?

katydecorah:
This PR switches the package manager from Yarn to npm. I pushed the latest commit to staging to confirm that the build succeeds.

@danswick for review 🙏 

Fixes #472


danswick (approved):
Yes! Looks great to me 🎉

PR #481, opened 2021-01-26
note that sky layer does not require source

domlet:
I noted [in Saman's PR](https://github.com/mapbox/studio-manual/pull/145/files#r564817088) the info that sky layers do not require a source.

This PR adds "or sky" here:

> Except for layers of the <var>background</var> or <var>sky</var> type, each layer needs to refer to a source. Layers take the data that they get from a source, optionally filter features, and then define how those features are styled.

Tagging @ryanhamley @HeyStenson for review

/cc: @samanpwbb


ryanhamley (approved):
Good catch!

HeyStenson (approved):
🍊

PR #491, opened 2021-02-12
Retitle GL JS example to "Add markers to a web map with a symbol layer"

domlet:
Re: JAMTAM https://github.com/mapbox/help/pull/2848

This example is our clearest example of using a symbol layer to draw points on a map; I am changing the title to make it recognizable to users of the ^ forthcoming JAMTAM getting started guide.

- draw points > add markers (titles can be slightly less technical, I think `markers` is a more user-centric term, and not inaccurate as this example uses markers)

Tagging @HeyStenson @ryanhamley for 👀 ✔️


domlet:
I also retitled `Add a default marker` to `Add a default marker to a web map` to add clarity for JAMTAM users.

ryanhamley (approved):

HeyStenson (approved):
🍊 👍 ✨

Issue #483, opened 2021-01-27
List required props (before optional props) in style spec


domlet:
## Motivation

- reduce user errors due to missing required props
- increase users' familiarity with vital props

## Design Alternatives

- props are currently listed alphabetically, which is sensible but could be improved

## Design

- list required props, alpha a-z
- list optional props, alpha a-z

### Mock-Up

on a page like this:
https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/

it would look like this:

<img src="https://user-images.githubusercontent.com/6026447/105925362-da0ee400-5ff4-11eb-8252-4636d6603292.jpg" width=300>

PR #492, opened 2021-02-12
Sorts Style spec properties by required then optional

katydecorah:
This PR sorts the Style spec properties by required and then optional properties. 

@domlet for review

Fix https://github.com/mapbox/mapbox-gl-js-docs/issues/483


domlet (approved):
Works beautifully – great improvement to surface required properties first, especially on `root` and `layers` style spec pages.

before/after Screenshot:

PR #493, opened 2021-02-16
Fix HTML clusters not always updating after moving the map

mourner:
Currently the [HTML clusters example](https://docs.mapbox.com/mapbox-gl-js/example/cluster-html/) uses `move` and `moveend` events to update clusters on the screen. However, this leads to subtle bugs when `moveend` happens before underlying tile data updates, since `moveend` is based on map movement and not tiles. 

You can reproduce this in Safari & Firefox (but weirdly not Chrome) clicking on the "zoom in" control button (clusters aren't updated), then dragging the map (they're updated when starting the movement instead).

This change makes the clusters get updated on every map render instead — this ensures the clusters on screen are always up to date, while not being slower than the current way because `move` and `render` event counts are roughly in the same ballpark (`render` just happens a few more times after movement ends).

Generally we should have a new event in GL JS that would be fired whenever underlying tile data changes — the `data` event currently doesn't cover many cases like tiles being removed, or added from cache instead of loading from source. I'll ticket this separately so that we can improve this example (and avoid marker updates on every frame) in the future.


allanwalkerit (approved):

allanwalkerit (approved):

allanwalkerit:
Thank you @mourner

mourner:
Apologies, there was a typo in the PR — followed up with a fix here https://github.com/mapbox/mapbox-gl-js-docs/commit/7b385e8402f2332c9942d54e1963f6c78db95c8d

PR #494, opened 2021-02-17
Replace dynamic iframes in examples

katydecorah:
This PR replaces `document.write()` to generate the iframe for examples with Webpack loaders. This change will improve performance of examples and remove `document.write()` which is [not a best practice](https://web.dev/no-document-write/).

## What's changed with examples?

Example markdown files will import its companion html file twice, one will have the [resource query](https://webpack.js.org/api/loaders/#thisresourcequery) `?code` and the other `?iframe`. These resource queries will allow Webpack loaders to generate the copiable code snippet (`?code`) and an HTML page to generate the demo iframe (`?iframe`).

The PR includes tests to assert that each example markdown file has the imported html with required resource queries and that the `Example` component has the `iframeSrc` prop.

```diff
prependJs:
- "import html from './simple-map.html';"
+- "import html from './simple-map.html?code';"
+- "import iframe from './simple-map.html?iframe';"
```

And then pass a new prop to the Example component:

```diff
- {{ <Example html={html} {...this.props} /> }}
+ {{ <Example html={html} iframeSrc={iframe} {...this.props} /> }}
```

## Full changes 

- Example copyable code and the iframe code is now generated by Batfish during start/build. The `example/*.html` files are still watched and will live reload during development. However, if you change the markup in `example-utils.js`, you may need to rerun Batfish to see changes reflected (not common).
- Combines `urls.js` function to `example-utils.js` and refactors slightly to make it compatible with all systems.
- Updates other files that used `urls.js` to use new function.
- Adds test to assert that each example's `prependJs` includes and `?code` and `?iframe` import.
- Adds test to assert that each example's content includes the `Example` component with the new `iframeSrc` prop.
- Refactors `Example` component to use pure component and pure function to improve performance.

This branch is live on staging.

Ref https://github.com/mapbox/documentation/issues/622


scripts/mapbox-shell-token.js
```diff
@@ -0,0 +1,16 @@
+const { readFileSync } = require('fs');
+
+// extract tokens from page-shell (this is a little fragile)
```
katydecorah, line 3: We use this technique in /help to get the production Mapbox token from the page shell. I don't love this technique as it feels brittle. Should we remove the file and add a direct token?

danswick: Mentioned in MTS, but the only concern i have with adding directly is thinking about how we'll update later.

docs/components/example-utils.js
```diff
@@ -0,0 +1,83 @@
...
+const css = `body { margin: 0; padding: 0; }
+#map { position: absolute; top: 0; bottom: 0; width: 100%; }`;
+const { version } = require('../../mapbox-gl-js/package.json');
+
+function urls(options) {
+    const prefix =
+        options && options.local && process.env.DEPLOY_ENV === 'local'
+            ? '/mapbox-gl-js/dist'
+            : `https://api.mapbox.com/mapbox-gl-js/v${version}`;
+    return {
```
danswick, line 19: this is a nice way to swap from the old js and css functions!

danswick (approved):
Looks great, @katydecorah! Tested locally as well and everything works well.

scripts/mapbox-shell-token.js
katydecorah: Ok! Let's keep as is. If this extract code gets funky, we could consider adding a json file to the docs-page-shell build process with tokens along with the page shell code to make these values easier to reach.

PR #496, opened 2021-02-25
Update Three.js example to improve state interaction

rreusser:
Three.js's `renderer.state.reset()` as presented in the current example leads to the following problematic WebGL state interactions when using Three.js in a custom layer:
- https://github.com/mapbox/mapbox-gl-js/issues/10372
- https://stackoverflow.com/questions/65707255/change-from-three-js-v122-to-v123-hides-the-shadows-on-a-custom-layer-for-mapbox
- https://github.com/mrdoob/three.js/pull/21281#issuecomment-779257362 (solution resulting from SO question)

[As explained in more detail here](https://github.com/mapbox/mapbox-gl-js/issues/10372#issuecomment-785615246), the most recent release of Three.js implements [`renderer.resetState()`](https://threejs.org/docs/index.html#api/en/renderers/WebGLRenderer.resetState) which alleviates the above problems.

I suspect GL-JS may also benefit from enforcing its state after calls to custom layers, and different layer orders may be a workaround, but this PR should hopefully at least start people off on the right foot with a minimal change that causes Three.js to render correctly.

A live version of the updated example: https://codepen.io/rsreusser/pen/abBEEXQ

cc @mapbox/gl-js


katydecorah (approved):
Works great, thanks @rreusser! 

I'll merge this in and the changes will be live in about 10 minutes.

PR #508, opened 2021-03-08
Exclude assets from sitemap

katydecorah:
This PR updates Batfish to use the new `ignoreFile` option for `sitemap`. We set `ignoreFile` to exclude all files in `assets` from being added to the sitemap (introduced in #494).


katydecorah:
The sitemap is live on staging! @danswick for review! No rush, anytime this week is great.

katydecorah:
@danswick do you think you could review this week?

danswick (approved):
This looks like it's working great, @katydecorah! I ran the static build locally just to be sure and it seems to be working beautifully.

PR #511, opened 2021-03-11
Update bathymetry tileset reference

millzpaugh:
We're writing a blog post about using [MTS to generate a bathymetric tileset ](https://github.com/mapbox/community-projects/issues/811)and we're updating this example to use the new MTS tileset!

### DIFF
--- a/docs/pages/example/style-ocean-depth-data.html
+++ b/docs/pages/example/style-ocean-depth-data.html
@@ -11,17 +11,17 @@
     });
 
     map.on('load', function () {
-        map.addSource('10m-bathymetry-81bsvj', {
+        map.addSource('10m-bathymetry', {
             type: 'vector',
-            url: 'mapbox://mapbox.9tm8dx88'
+            url: 'mapbox://examples.bathy_z8'
         });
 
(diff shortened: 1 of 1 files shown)


katydecorah:
@millzpaugh the changes in this branch look very different from what's on publisher-production:

publisher-production | current branch
---|---
image | image

This error is logged to the console:

```
Error: Source layer "water-detail-v1" does not exist on source "10m-bathymetry" as specified by style layer "10m-bathymetry-id"
    at i._validateLayer (mapboxgl.js:1)
    at i.<anonymous> (mapboxgl.js:1)
    at i.Ft.fire (mapboxgl.js:1)
    at i.Ft.fire (mapboxgl.js:1)
    at i.Ft.fire (mapboxgl.js:1)
    at mapboxgl.js:1
    at r (mapboxgl.js:1)
```

colleenmcginnis:
@millzpaugh what's the status of this example? Note: after you resolve the issue Katy referenced above, you should also provide some context about using custom tilesets similar to what is done in https://github.com/mapbox/mapbox-gl-js-docs/pull/535.

Issue #16, opened 2019-08-09
Prevent second click for /mapbox-gl-js/example/toggle-layers/


danswick:
*This feedback was submitted by a user through the feedback widget on the bottom of the page.*

* Page: [/mapbox-gl-js/example/toggle-layers/](https://docs.mapbox.com/mapbox-gl-js/example/toggle-layers/)
* Section: null
* Date received: 2019-08-06T10:33:17.687Z
* Was this page helpful?: 😊

Feedback:
> This prevents having to click a second time in some browsers:
if (visibility === 

We'll need to check this out in a few browsers to see where the issue might be happening.

andrewharvey:
I suspect this is when you click on the button before the map and layer have loaded. A better solution is to have these buttons disabled until the layers have been added to the map.

PR #516, opened 2021-03-15
Wait to enable buttons until layers exist in Show and hide layers example

colleenmcginnis:
Fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/16

Wait to enable the toggle buttons until the `contours` and `museums` layers have been added in the [Show and hide layers](https://docs.mapbox.com/mapbox-gl-js/example/toggle-layers/) example.


docs/pages/example/toggle-layers.html
```diff
@@ -95,38 +140,15 @@
                 'line-width': 1
             }
         });
+        // After the layer is added to the style, enable the
+        // associated toggle item.
+        if (map.getLayer('contours')) enableToggleItem('contours');
+        if (map.getLayer('museums')) enableToggleItem('museums');
```
ryanhamley, line 146: I think this would work mostly by coincidence. It should work but I don't think it's guaranteed; if the layer isn't done parsing by the time these lines are called, this is never called again and the toggles will never be enabled. I think if you move this out of the `load` function and put it into an `idle` function, it would be more reliable.

You could also think about adding the entire toggle control in the idle event rather than enabling it separately from creation.

ryanhamley (changes requested):

docs/pages/example/toggle-layers.html
colleenmcginnis: Thanks @ryanhamley. I've actually never used `idle` before. Is this what you had in mind? https://github.com/mapbox/mapbox-gl-js-docs/pull/516/commits/f6f62e7cba4f235cc4bc96d2de062ca063841f2f

ryanhamley: Yep! Although you could simplify the code further by not breaking the functionality out into a separate function.

`if (map.getLayer('contours') && map.getLayer('museums')) buildToggle();` could become

```
if (map.getLayer('contours') && map.getLayer('museums')) {
    // contents of buildToggle
}
```

ryanhamley (approved):

PR #517, opened 2021-03-15
Add `workerClass` to documentation.yml

katydecorah:
This PR adds `workerClass` to documentation.yml so that it will appear on https://docs.mapbox.com/mapbox-gl-js/api/properties/

> image

Fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/515


colleenmcginnis (approved):
Thanks @katydecorah! 

cc @arindam1993 (As far as I can tell this was introduced in https://github.com/mapbox/mapbox-gl-js/pull/10219. Should there be something in the "Launch Checklist" in `mapbox-gl-js` to `docs/documentation.yml` if adding something new?)

arindam1993:
Ahh I was not aware that a manual entry was required, yes we should add this to the Launch Checklist.

Issue #64, opened 2019-09-07
It is difficult to find information on how to link to a custom tileset


danswick:
*This feedback was submitted by a user through the feedback widget on the bottom of the page.*

* Page: [/mapbox-gl-js/example/vector-source/](https://docs.mapbox.com/mapbox-gl-js/example/vector-source/)
* Section: null
* Date received: 2019-09-02T23:21:53.583Z
* Was this page helpful?: 😕

Feedback:
> It is difficult to find information on how to link to a custom tileset.  This example uses a link to a mapbox tileset. I assume that all mapbox ileset have a URL, but how do I find the URL I am suppose to use? And what URL do I use for custom tilesets? How do should I use the tileset ID? 
>
>You would think custom tilesets would be a big part of what people need from mapbox, and yet, I feel like there is very little documentation on how to actually do that.  

I support this idea! Interoperability is a big part of what makes mapbox distinct.

PR #518, opened 2021-03-15
Add link to tileset id definition in Add a vector tile source example

colleenmcginnis:
Fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/64

Adds a link to the glossary entry for "tileset id" to help the reader better understand what it means to add a vector source in the [Add a vector tile source](https://docs.mapbox.com/mapbox-gl-js/example/vector-source/) example.


ryanhamley (commented):
I think this helps a bit, but it doesn't really address what seem like the central concerns of the feedback.

> I assume that all mapbox ileset have a URL, but how do I find the URL I am suppose to use?

Is there anywhere that lists all available mapbox tilesets and their URLs so people can easily get the URL? They're referenced all over the place but I don't think there's a place that documents all of them that I know of.

> And what URL do I use for custom tilesets? 

We could be more explicit about getting them from Studio and/or hosting your own WMTS.

> How do should I use the tileset ID?

I don't know about this part. It's pretty basic. Maybe being explicit that the tileset ID is part of the URL and again focusing on where to find the URL?

I think the fuller solution here may go beyond changes just to the GL JS docs.

colleenmcginnis:
Thanks for the thoughtful comments @ryanhamley! What do you think about this additional context? Signing into Studio may be more involved than just listing tilesets provided by Mapbox on a docs page, but I think it's the right place to point users since it includes _all_ tilesets they can use with their account (default Mapbox ones like streets and custom tilesets they've uploaded to their account).

Also, I know that users _can_ also use non-Mapbox-hosted tilesets, but I'm not sure we want to get into that much detail on an example page (unless we receive additional feedback asking for it).

image

ryanhamley (approved):
Looks great! I think that's a helpful change

Issue #7, opened 2019-08-05
where do you load sprites


katydecorah:
*This feedback was submitted by a user through the feedback widget on the bottom of the page.*

* Page: [/mapbox-gl-js/example/geojson-markers/](https://docs.mapbox.com/mapbox-gl-js/example/geojson-markers/)
* Section: null
* Date received: 2019-08-01T22:28:16.606Z
* Was this page helpful?: 😕

Feedback:
> where do you load sprites

---

It isn't clear where the icon images are referenced. We should add additional context to the page's description.

PR #522, opened 2021-03-17
Add context when referencing icons in a style's sprite

colleenmcginnis:
Fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/7
Fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/180
Fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/249

Many users have provided feedback on examples that reference icons that exist in Mapbox Streets (or other template styles) wondering where they can find a list of icon names that are available in a style. Based on input from members of the @mapbox/map-design-team, it sounds like we shouldn't be encouraging users to rely on icons that are included by default in template styles. Instead we should start encouraging more use of the run `setImage` API method in our examples in cases where users want to display icons or patterns alongside client side data.

This PR updates examples that currently reference icons from Mapbox Streets including:

Uses one image:
- [x] https://docs.mapbox.com/mapbox-gl-js/example/animate-point-along-route/
- [x] https://docs.mapbox.com/mapbox-gl-js/example/camera-animation/
- [x] https://docs.mapbox.com/mapbox-gl-js/example/center-on-symbol/
- [x] https://docs.mapbox.com/mapbox-gl-js/example/change-case-of-labels/
- [x] https://docs.mapbox.com/mapbox-gl-js/example/filter-features-within-map-view/
- [x] https://docs.mapbox.com/mapbox-gl-js/example/live-geojson/
- [x] https://docs.mapbox.com/mapbox-gl-js/example/popup-on-hover/

Uses multiple images:
- [x] https://docs.mapbox.com/mapbox-gl-js/example/fallback-image/
- [x] https://docs.mapbox.com/mapbox-gl-js/example/filter-markers-by-input/
- [x] https://docs.mapbox.com/mapbox-gl-js/example/filter-markers/
- [x] https://docs.mapbox.com/mapbox-gl-js/example/popup-on-click/
- [x] https://docs.mapbox.com/mapbox-gl-js/example/variable-label-placement/

In most cases I just used a png of the Maki icon that was used before (but usually a little bigger), but if we think that is confusing we can replace them with different icons (if anyone has recommendations for open source icons let me know!).

@mapbox/gl-js a few questions:
- [x] Does this approach sound ok to you?
- [x] ~~Is there a good way to add multiple images at runtime (I was thinking of using a promise, but open to other suggestions)?~~
- [x] ~~Is there any benefit to including both `addSource` and `addLayer` in the `loadImage` callback or is it ok to do `addSource` outside of it?~~

Still to do: 
- [x] ~~Update image references to use absolute URLs before merging.~~


ryanhamley (commented):
I'm a little unsure of this approach. It adds a decent chunk of code to implement something that already exists. It could be somewhat confusing and it makes GL JS seems slightly less functional out of the box. People might assume they always need to come up with their own icons.

I can understand the map-design team's position. How often do the icons change? They don't guarantee icon names for a given version of a style?

asheemmamoowala:
> Based on input from members of the @mapbox/map-design-team, it sounds like we shouldn't be encouraging users to rely on icons that are included by default in template styles. Instead we should start encouraging more use of the run setImage API method in our examples in cases where users want to display icons or patterns alongside client side data.

This depends. `setImage` is great for dynamic resources. Adding images to sprites defined by the style ensure that the image is available sooner and can be used in symbol layers defined in the style and modified using runtime styling. 

Would it make sense to create a new examples style based on a Mapbox style with added icons specific to these examples, and link to a tutorial on adding images to a style's sprite using Studio?

docs/pages/example/animate-point-along-route.md
```diff
@@ -14,8 +14,15 @@ prependJs:
...
+- "import Note from '@mapbox/dr-ui/note';"
 ---
 
 Use [Turf](http://turfjs.org/) to smoothly animate a point along the distance of a line.
 
+{{<Note title="Referencing images in a style's sprite">}}
+This example uses the [Mapbox Streets](https://www.mapbox.com/maps/streets) style. The airplane `icon-image` used in the `point` layer comes from the Mapbox Streets style's [sprite](https://docs.mapbox.com/help/glossary/sprite/). 
+
+To view all available images in a style's sprite or add additional images, open the style in [Mapbox Studio](https://studio.mapbox.com/styles/) and click the "Images" tab. To add a new image to the style at runtime see the [Add an icon to the map](https://docs.mapbox.com/mapbox-gl-js/example/add-image/).
+{{</Note>}}
```
colleenmcginnis, line 26: Ok the intention here is to shed some light on where `airport-15` is coming from so the user at least has a lead if they want to know what icons they can use out-of-the-box.

Is this too much information? Is suggesting users open a style in Mapbox Studio too much to ask? We don't document a list of all available icons per style and the Map design team has recommended that we don't.

docs/pages/example/center-on-feature.html
```diff
@@ -0,0 +1,83 @@
...
+            'id': 'circle',
+            'type': 'circle',
+            'source': 'points',
+            'paint': {
+                'circle-color': '#4264fb',
+                'circle-radius': 8,
+                'circle-stroke-width': 2,
+                'circle-stroke-color': '#ffffff'
+            }
+        });
```
colleenmcginnis, line 64: For most of the other examples, I don't think using a symbol layer is critical to the concepts we're trying to illustrate (like centering on a feature when it's clicked). When this is the case, what do you think about just using a circle layer so we don't have to muddy the concept we're trying to illustrate with information about how to find available images in a style's sprite.

docs/pages/example/live-geojson.html
```diff
@@ -32,13 +32,17 @@
...
+            if (error) throw error;
+            map.addImage('rocket-medium', image);
+            map.addLayer({
+                'id': 'drone',
+                'type': 'symbol',
+                'source': 'drone',
+                'layout': {
+                    'icon-image': 'rocket-medium'
+                }
+            });
```
colleenmcginnis, line 45: I didn't update this one, but I image it would require similar treatment to `docs/pages/example/animate-point-along-route.html` above.

colleenmcginnis (commented):
A couple ideas below. Let me know what you think.

docs/pages/example/animate-point-along-route.md
colleenmcginnis: The note is a bit overwhelming, but it is a complex thing to explain. 

image

ryanhamley: I think this is great. It's very clear and offers a lot of options for learning more about the concepts of icons and sprites and how to use them.

ryanhamley (commented):
I think all the changes here are fine for what this PR is trying to achieve. My question though is if this is a bit overkill. Is this such a frequent area of confusion that we have to explicitly call out where the symbols come from in every example and avoid using them whenever possible? The `popup-on-click` and `popup-on-hover` examples make more sense with icons denoting the type of entertainment at each location rather than just a simple circle, for example.

What if we just had one simple example of using symbols in a style that had the explicit callouts? Even if you still think it's necessary to reference what's going on in every single example, we could link to that one example rather than copying and pasting the same text all over the place (which also adds a lot of overhead if we want to change the wording).

I also still think this change to circles instead of symbols might go too far in the opposite direction and make GL JS seem less capable than it is while being confusing in its own right. The natural question a lot of developers might have is "OK, but what if I want an icon instead of a circle?" A quick glance at some of these examples may make it seem like that's not possible or difficult.

Thoughts? I'm definitely open to persuasion.

colleenmcginnis:
Thanks for these questions @ryanhamley! 

>Is this such a frequent area of confusion that we have to explicitly call out where the symbols come from in every example and avoid using them whenever possible?

@mapbox/docs and @mapbox/support may have a different opinion here, but based on my experience this is one of _the most_ consistent lines of user feedback that we've seen coming through the "was this helpful" component.

>What if we just had one simple example of using symbols in a style that had the explicit callouts? Even if you still think it's necessary to reference what's going on in every single example, we could link to that one example rather than copying and pasting the same text all over the place (which also adds a lot of overhead if we want to change the wording).

I think wherever possible we avoid creating examples that require visiting and understanding another example in order to fully understand how the first one works. This pattern goes against one of the best practices our team follows to treat "every page as page one" (a reader can land on any page on your site from anywhere, so each individual page should exist as a freestanding, complete piece).

**But**, I do understand your concern about maintenance. Another approach we could take is to create a React component with the contents of the note and just load it on pages that reference icons in a style's sprite. That way the information is on every relevant page, but only needs to be maintained in one place. 

>The natural question a lot of developers might have is "OK, but what if I want an icon instead of a circle?"

I'm sure this is true, but we also get feedback in the other direction (i.e. this shows icons but I want circles... the grass is always greener etc 😛). Our catalog of examples could quickly become unmanageable if we acted on every piece of that kind of feedback. Our approach is to make sure each example is properly scoped to illustrate how a single feature/API or end result is achieved.

For example, the part of the code that shows "how to add a pop-up to a feature on hover" works the same for features in any circle, fill, line, or symbol layer. We could create a separate example for each layer type or we could make the assumption that the reader can copy-paste(-and-experiment) on their own to make it work for the type of feature that is most appropriate for their use case.

But, if any of the examples I changed to circles are really better illustrated by symbols we can and should use symbols! 

@mapbox/docs curious if you have additional thoughts here.

ryanhamley:
>I think wherever possible we avoid creating examples that require visiting and understanding another example in order to fully understand how the first one works. This pattern goes against one of the best practices our team follows to treat "every page as page one" (a reader can land on any page on your site from anywhere, so each individual page should exist as a freestanding, complete piece).
> 
> **But**, I do understand your concern about maintenance. Another approach we could take is to create a React component with the contents of the note and just load it on pages that reference icons in a style's sprite. That way the information is on every relevant page, but only needs to be maintained in one place.

OK, I'm on board with the "page one" approach. That makes total sense. I think a React component that reduces maintenance would be a great compromise solution.
 
> I'm sure this is true, but we also get feedback in the other direction (i.e. this shows icons but I want circles... the grass is always greener etc 😛). Our catalog of examples could quickly become unmanageable if we acted on every piece of that kind of feedback. Our approach is to make sure each example is properly scoped to illustrate how a single feature/API or end result is achieved.
> 
> For example, the part of the code that shows "how to add a pop-up to a feature on hover" works the same for features in any circle, fill, line, or symbol layer. We could create a separate example for each layer type or we could make the assumption that the reader can copy-paste(-and-experiment) on their own to make it work for the type of feature that is most appropriate for their use case.
> 
> But, if any of the examples I changed to circles are really better illustrated by symbols we can and should use symbols!

👍 The main reason I said that symbols might work better for the popup on hover/click examples was the context they provide within that specific example (ie "this thing is a music venue; click to get more info on it"), but it's not strictly necessary as you say. I'd definitely lean towards simplifying the docs and trusting users to experiment and copy/paste.

I'd say this is a solid change then and just suggest the React component to make maintenance easier in the future.

colleenmcginnis:
@ryanhamley I pushed a few updates: 

1️⃣  https://github.com/mapbox/mapbox-gl-js-docs/pull/522/commits/95732b05132a8a7b8a50c473fe683d2c6f57348e: Changed the `popup-on-click` example back to using a symbol layer (per your comment). IMO including the full icon name in the GeoJSON `icon` property (i.e. with the `-15`) is easier to digest.
2️⃣  https://github.com/mapbox/mapbox-gl-js-docs/pull/522/commits/e53fd54e1d8fe3ec59bae37ae575c36bbeeb0eb0: Created a note component that can be passed a style name in each Markdown file. The result looks like this:

Screen Shot 2021-04-12 at 10 22 37 AM

Screen Shot 2021-04-12 at 10 22 16 AM

ryanhamley (approved):
This looks great! Thanks for bearing with me

PR #524, opened 2021-03-21
Update hover-styles.html

mvl22:
The example fails when an ID of a feature is zero (0), such that the hover state does not clear. This fix does a strict check against exactly null (matching what is set in lines 9 and 74) rather than a falsey check.


katydecorah (approved):
Thank you!

PR #528, opened 2021-03-25
Update GL JS to version 2.2.0

ansis:
Update GL JS to version 2.2.0

@mapbox/docs


mourner (changes requested):
Looks like the tests failed because `api-navigation` test snapshot has to be updated with new items?

katydecorah (approved):

PR #531, opened 2021-03-31
Set polygon draw example to polygon_draw mode by default. Update title and description.

danswick:
Fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/196

Updates the "Show drawn polygon area" title and description to better reflect what the example is showing. Also uses mapbox-gl-draw's "modes" feature to enable drawing by default without the user having to click on the polygon button first.

```js
var draw = new MapboxDraw({
  displayControlsDefault: false,
  controls: {
    polygon: true,
    trash: true
  },
  defaultMode: 'draw_polygon'
});
```

image


danswick:
@mapbox/docs - this example is kind of straddling two purposes right now: showing off Turf and showing off mapbox-gl-draw. Do you think it's ok this way or should it tilt more in one way or another?

docs/pages/example/mapbox-gl-draw.md
```diff
@@ -16,6 +16,6 @@ prependJs:
 - "import iframe from './mapbox-gl-draw.html?iframe';"
 ---
 
-Use [mapbox-gl-draw](https://github.com/mapbox/mapbox-gl-draw) to draw a polygon and [Turf.js](http://turfjs.org/) to calculate its area in square meters.
+Click the map to begin drawing a polygon. As you add vertices to your polygon, its total area will appear in the lower left of the map. This examples uses [mapbox-gl-draw](https://github.com/mapbox/mapbox-gl-draw) to draw the polygon and [Turf.js](http://turfjs.org/) to calculate its area in square meters.
```
colleenmcginnis, line 19: There might be a better way to word this, but I do think readers are often confused about how to stop drawing once you've started...

```suggestion
Click the map to begin drawing a polygon. As you add vertices to your polygon, its total area will appear in the lower left of the map. Double-click when you draw the final vertex to complete the polygon. This examples uses [mapbox-gl-draw](https://github.com/mapbox/mapbox-gl-draw) to draw the polygon and [Turf.js](http://turfjs.org/) to calculate its area in square meters.
```

colleenmcginnis (approved):

domlet:
> this example is kind of straddling two purposes right now: showing off Turf and showing off mapbox-gl-draw. Do you think it's ok this way or should it tilt more in one way or another?

If we have no user or Support team complaints, I don't mind the hybridization. I think it's a cool demo app that gains interestingness for what it sacrifices in single-purposeness. I think it could benefit from a more detailed description and code comments that would equip a user learning to import libraries/plugins to better understand which thing is doing what.

danswick:
Thanks, @colleenmcginnis and @domlet for your input and reviews! I am pretty happy with these changes for now and think it should give our users a better experience with this example. If we hear more feedback, we can revisit!

PR #533, opened 2021-04-02
Improve 2 of 20 most-visited GL JS Examples

domlet:
Re: https://github.com/mapbox/documentation/issues/654#

* Adjusts example content to improve visuals
* Elaborates on descriptions in plain language
* Adds links to API/Style Spec reference docs when mentioned 
* Adds code comments to increase accessibility
* Adds trailing slashes to links in this repo

## Before/after screenshots:

* `docs/pages/example/add-image.md`
* `docs/pages/example/geojson-polygon.md`

Screen Shot 2021-04-02 at 9 48 31 AM

PR #536, opened 2021-04-05
Set `width` and `height` on Layers page images to prevent layout shift

katydecorah:
This PR fixes a layout shift on the Layers page as described in https://github.com/mapbox/mapbox-gl-js-docs/pull/533.

Due to a limitation with AppropriateImages, I had to make these images background images to be able to set the width and height attributes - there should be no difference in how users are able to view these images. I also made an adjustment to the Caption component so that it has the same width as the image it is captioning.

@domlet for review.


domlet (approved):
Thank you so much @katydecorah! 🙏

PR #537, opened 2021-04-07
Create a GL JS example that loads an external geojson file

domlet:
Closes https://github.com/mapbox/mapbox-gl-js-docs/issues/297

* Adds an example that loads a GeoJSON file from an external URL
* Re-uses existing `/assets/earthquakes.geojson` resource
* Limits data in another example to six-digits: `geojson-line.html`
* Reduces zoom in another example (so markers appear on page load even when user screen is smaller): `geojson-markers.html`

In this example, we use a file on our own domain, but demonstrate the pattern that developers can use for external URLs, CORS willing.

* Similar to https://docs.mapbox.com/mapbox-gl-js/example/live-geojson/
* Different from https://docs.mapbox.com/mapbox-gl-js/example/free-camera-path/


domlet:
@danswick Are there advantages of offering one of these two patterns over the other for this example?

## 1

```
var url = 'https://docs.mapbox.com/mapbox-gl-js/assets/earthquakes.geojson';
 
map.on('load', function () {
map.addSource('earthquakes', {
type: 'geojson',
data: url
});
```

## 2 

```
map.on('load', function () {
map.addSource('earthquakes', {
type: 'geojson',
data: 'https://docs.mapbox.com/mapbox-gl-js/assets/earthquakes.geojson'
});
```

docs/pages/example/geojson-markers.html
```diff
@@ -4,7 +4,7 @@
         container: 'map',
         style: 'mapbox://styles/mapbox/light-v10',
         center: [-96, 37.8],
-        zoom: 3
+        zoom: 2
```
danswick, line 7: Is this an accidental change?

docs/pages/example/external-geojson.html
```diff
@@ -0,0 +1,34 @@
...
+<script>
+    var map = new mapboxgl.Map({
+        container: 'map', // container ID
+        style: 'mapbox://styles/mapbox/satellite-v9', // style URL
+        zoom: 7, // starting zoom
+        center: [138.043, 35.201] // starting center
+    });
+
+    // Define a variable for external URL string.
+    var url = 'https://docs.mapbox.com/mapbox-gl-js/assets/earthquakes.geojson';
```
danswick, line 12: I think you could go either way. Defining a variable makes sense if you think future maintainers may need to find and quickly change this value. For this example, I might opt for putting the string directly in the `data` property for the source. Since that property can be one of either be a URL string or raw GeoJSON, putting the URL there sort of reinforces what type of value you can use. Kind of a minor distinction, though, and I could see the merits of either choice!

docs/pages/example/external-geojson.html
```diff
@@ -0,0 +1,34 @@
...
+
+    map.on('load', function () {
+        // Use the variable for the 'data' value
+        // when you add the source to the map.
+        map.addSource('earthquakes', {
+            type: 'geojson',
+            data: url
+        });
```
danswick, line 20: If you do swap this, you could do something like this to retain the helpful comment from above:

```suggestion
            type: 'geojson',
            // Define a variable for external URL string.
            data: 'https://docs.mapbox.com/mapbox-gl-js/assets/earthquakes.geojson'
        });
```

danswick (commented):
Thanks for taking this on, @domlet! Added a couple of comments inline.

docs/pages/example/geojson-markers.html
domlet: @danswick Thanks for checking. No, this change from `3` to `2` ensures that both markers appear on the map when the example page first loads, even if the viewport is narrow.

docs/pages/example/external-geojson.html
domlet: Yes, def! Change made in 6753fef

docs/pages/example/external-geojson.html
domlet: Thanks @danswick – I removed the `var` abstraction and stuck the URL directly in `addSource` to make that connection very clear:

> putting the URL there sort of reinforces what type of value you can use

Followed your advice also on the code comment. Thank you!

docs/pages/example/geojson-markers.html
danswick: Ah ok. In general, it's good practice to a PR's changes pretty tightly scoped, but this is small enough that it's ok!

danswick (approved):
Looks great to me, @domlet!

domlet:
Thanks @danswick! 

Tagging @mapbox/gl-js for review 👍

docs/pages/example/external-geojson.html
```diff
@@ -0,0 +1,31 @@
...
+        zoom: 7, // starting zoom
+        center: [138.043, 35.201] // starting center
+    });
+
+    map.on('load', function () {
+        map.addSource('earthquakes', {
+            type: 'geojson',
+            // Use a URL for the value for the `data` property.
+            data:
+                'https://docs.mapbox.com/mapbox-gl-js/assets/earthquakes.geojson'
```
ryanhamley, line 16: nit: this doesn't need to be on a new line

ryanhamley (approved):
>Are there advantages of offering one of these two patterns over the other for this example?

There's really no advantage to option 1 unless you're going to use the `url` variable elsewhere in your code. If the only place the string is being used, I'd use option 2.

Looks good! Just move the url onto the same line as `data`.

docs/pages/example/external-geojson.html
domlet, line 16:
```suggestion
            data: 'https://docs.mapbox.com/mapbox-gl-js/assets/earthquakes.geojson'
```

domlet: 👍  thanks for spotting that! resolved in 9c62e65

docs/pages/example/geojson-markers.html
domlet: Good note; I'll try to avoid pork-barreling. Thanks Dan!

PR #542, opened 2021-04-21
Add more context to glyph section intro to clarify that users probably don't need to touch it

danswick:
fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/329

Adds some more context to the glyph page's intro paragraph to clarify that users probably won't ever have to worry about their style's glyphs property. There may be some users who need to run their own font server, but I don't think we should describe that process in the style spec reference.


docs/pages/style-spec/glyphs.md
```diff
@@ -13,7 +13,7 @@ prependJs:
 
 <!--copyeditor disable basic-->
 
-A style's `glyphs` property provides a URL template for loading signed-distance-field glyph sets in PBF format.
+A style's `glyphs` property provides a URL template for loading signed-distance-field glyph sets in PBF format. This property is created by Mapbox Studio automatically it is unlikely you will need to change it. Your map client (Mapbox GL JS, for example) will generate and send glyph requests to the specified font server using the URL template you provide.
```
HeyStenson, line 16:
```suggestion
A style's `glyphs` property provides a URL template for loading signed-distance-field glyph sets in PBF format. This property is created by Mapbox Studio automatically, and it is unlikely you will need to change it. Your map client (Mapbox GL JS, for example) will generate and send glyph requests to the specified font server using the URL template you provide.
```

HeyStenson (approved):
Beaut! Left one suggestion.

PR #543, opened 2021-04-21
[example] Reset features filter once the map starts moving

katydecorah:
This PR fixes an issue with https://docs.mapbox.com/mapbox-gl-js/example/filter-features-within-map-view/ where it was not possible to reset the map after you enter a search with the input. By reseting the filter on `movestart`, the map will reset in time for the map to keep the queried features and list in sync.

@danswick would you mind taking a look?

Fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/293


danswick (approved):
This looks like a solid change to me. In the future, we might consider getting ride of the text filter just to clarify the functionality in this example a bit, but this seems like should handily resolve the issues we've been seeing! 

I wasn't sure if this was still supposed to be a draft PR, so I'm going with an "official" review just in case :)

PR #552, opened 2021-04-28
Update style spec to 13.9.1 and update lock file

asheemmamoowala:
Updates to the latest style spec package to include compatibility table for `terrain`


HeyStenson (approved):
👍 🎉 

Looks good @asheemmamoowala! I rebuilt the lockfile with npm6. Since we haven't migrated to npm7 yet, we haven't determined whether there are any unexpected downstream effects. cc @katydecorah and @danswick

asheemmamoowala:
13.9.1 release was missing some files 😢

